PR #1, opened 2023-03-06
Renamed things from fabric to warp and wiped the changelog

magnuh:


AnnaRybkina (approved):
Nicely done 🥇 

Found a couple of other `fabric` mentions 👇 
https://github.com/warp-ds/react/blob/main/esbuild.mjs#L43
A couple of those in https://github.com/fabric-ds/react/blob/next/index.html
https://github.com/fabric-ds/react/blob/next/package.json#L103
https://github.com/fabric-ds/react/blob/next/package.json#L104
couple of those -> https://github.com/fabric-ds/react/blob/next/%40types/index.d.ts 
and more places but I guess we can fix them in upcoming PR:s

magnuh:
> Nicely done 🥇
> 
> Found a couple of other `fabric` mentions 👇 https://github.com/warp-ds/react/blob/main/esbuild.mjs#L43 A couple of those in https://github.com/fabric-ds/react/blob/next/index.html https://github.com/fabric-ds/react/blob/next/package.json#L103 https://github.com/fabric-ds/react/blob/next/package.json#L104 couple of those -> https://github.com/fabric-ds/react/blob/next/%40types/index.d.ts and more places but I guess we can fix them in upcoming PR:s

Changed a few more. 🙂  ...but a lot of them are imports which just won't translate directly to warp... so they need to be changed when there are new packages available in warp. 👍

PR #2, opened 2023-03-14
chore: Initial release setup

magnuh:
Changes introduced for the release are based on [our internal guide on adding automatic releases to Warp packages](https://github.com/warp-ds/release-playground/wiki/Automate-releases-with-semantic-release):
trigger a release with semantic-release whenever a commit is pushed to alpha or main branch.

Co-authored by @AnnaRybkina  and @BalbinaK


magnuh:
[Automate releases of warp packages to NPM and EIK](https://trello.com/c/v1Ptqg8G/128-automate-releases-of-warp-packages-to-npm-and-eik)

magnuh:
[Automate releases of warp packages to NPM and EIK](https://trello.com/c/v1Ptqg8G/128-automate-releases-of-warp-packages-to-npm-and-eik)

.github/workflows/release.yml
```diff
@@ -0,0 +1,29 @@
+name: Release
+on:
+  push:
+    branches:
+      - master
```
BalbinaK, line 5: main

.github/workflows/release.yml
```diff
@@ -0,0 +1,29 @@
...
+      - alpha
```
BalbinaK, line 6: Not something for this PR but I'm thinking that maybe we should add `beta` and `next` here in case we decide to name the prerelease branch something else, like they did in [eik plugin for semantic-release](https://github.com/eik-lib/semantic-release/blob/master/.github/workflows/publish.yml#L7-L9)
Then we'd also need to tell semantic-release that those branches can be treated as a prerelease in `.releaserc.json`.
What do you think?

.releaserc.json
```diff
@@ -0,0 +1,21 @@
...
+    ],
+    [
+      "@semantic-release/github",
+      {
+        "assets": "release/*.tgz"
+      }
+    ],
+    "@semantic-release/git"
+  ]
+}
```
BalbinaK, line 21: Since we have a `.releaserc.json` file, the `release.config.js` file can be removed.

BalbinaK (commented):
Just a few small comments 😊

magnuh:
> Just a few small comments 😊

thanks 🙏  fixed 🙂

.github/workflows/pr-title-linter.yml
```diff
@@ -0,0 +1,105 @@
+name: PR Title Checker
```
BalbinaK, line 1: We can now avoid copy-pasting this logic and use a github action instead like [here](https://github.com/warp-ds/component-classes/pull/7#pullrequestreview-1341058736)

magnuh: Yup, done. 🙂

.releaserc.json
```diff
@@ -0,0 +1,21 @@
+{
+  "branches": [{ "name": "main" }, { "name": "alpha", "prerelease": true }],
```
BalbinaK, line 2: Since you added other branches to the .github/workflows/release.yml, we need to declare them as prerelease here, too :)

BalbinaK (approved):

PR #5, opened 2023-03-22
fix: Styling alert component with tokens

magnuh:
Known issues / open questions: 

- We need classes for styling color of individual borders, like `border-l` to be able to do: `i-border-l-$color-alert-success-border`. (Did this through the style attribute for now, as in Fabric/react.)
- The automatic class factory doesn't work when trying to dynamically create the classes like using a prop in react for setting the type `i-border-$color-alert-${type}-subtle-border`...but it DO work, however, once I have used the class name hard coded before doing it dynamically, like `i-border-$color-alert-success-subtle-border`. Guessing that the css compiler has already run when the class gets set in react... 🤷 (This thing was driving me crazy, when some things worked and some not. 😅🙈)
- Still left to decide how the css theme file should be included... I just added it as a `<link>` in `<head>` for now.
- The non-dynamic class list of this component should go into the component-classes repo, right?
- The order of the different parts of the token seems somewhat messed up for the alert component (or the others 😅): Should it be `w-alert-color` or `w-color-alert`? (button is `w-button-color`)
- The icons should be themeable as well, right?

image

### DIFF
--- a/index.html
+++ b/index.html
@@ -17,6 +17,7 @@
         visibility: visible;
       }
     </style>
+    <link rel="stylesheet" href="https://assets.finn.no/pkg/@warp-ds/tokens/1.0.0-alpha.9/finn-no.css" />
   </head>
   <body>
     <div class="m-16">
--- a/index.html
+++ b/index.html
@@ -9,15 +9,7 @@
       href="https://www.finn.no/favicon.ico"
       type="image/x-icon"
     />
-    <style>
-      f-docs-template {
-        visibility: hidden;
-      }
-      f-docs-template:defined {
-        visibility: visible;
-      }
-    </style>
-    <link rel="stylesheet" href="https://assets.finn.no/pkg/@warp-ds/tokens/1.0.0-alpha.9/finn-no.css" />
+    <link rel="stylesheet" href="https://assets.finn.no/pkg/@warp-ds/tokens/v1/finn-no.css" />
   </head>
   <body>
     <div class="m-16">
--- a/index.html
+++ b/index.html
@@ -17,29 +17,9 @@
         visibility: visible;
       }
     </style>
-    <link
-      href="https://assets.finn.no/pkg/@fabric-ds/css/v1/fabric.min.css"
-      rel="stylesheet"
-    />
-    <link
-      href="https://assets.finn.no/pkg/@fabric-ds/common/v1/css/markdown.css"
-      rel="stylesheet"
-    />
-    <link
-      href="https://assets.finn.no/pkg/@fabric-ds/common/v1/template/index.js"
-      rel="modulepreload"
-    />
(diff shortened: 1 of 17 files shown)


index.html
```diff
@@ -17,6 +17,7 @@
         visibility: visible;
       }
     </style>
+    <link rel="stylesheet" href="https://assets.finn.no/pkg/@warp-ds/tokens/1.0.0-alpha.9/finn-no.css" />
```
pearofducks, line 20: This should use an Eik alias so it doesn't need constant updates

pearofducks:
> The non-dynamic class list of this component should go into the component-classes repo, right?

Yup!

index.html
magnuh: Yup! 👍 Don't think that there was an alias... and not sure whether this link was the way to go forward, so just included it to get started. 😄

magnuh:
> I don't think we want to change the API (e.g. `negative` -> `critical`) unless we have to, especially when it's something that is just a 'wording change' and isn't really meaningful IMO.
> 
> I don't know if this means the names need aligned back with design (e.g. these names would show up in Figma, and thus confusing developers because there's a difference) or if these are just token names no one but Adi will ever see.

I think the new names are better... but if that will be a problem for migrating or something else then it might be better to change it in the design as well... Less confusing to use the same words across all domains. I'll check with Adi. 👍

magnuh:
> Yup! (I assume you mean they should have a color token for them we should be using)

The color tokens for the icons are already in place. 🙂  But I thought that the icons themselves could be different on all brands? Or are we going to use the same svg for all?

index.html
magnuh: Alias in place now, so I updated this. 👍

index.html
```diff
@@ -9,14 +9,7 @@
...
     />
-    <style>
-      f-docs-template {
-        visibility: hidden;
-      }
-      f-docs-template:defined {
-        visibility: visible;
-      }
-    </style>
+    <link rel="stylesheet" href="https://assets.finn.no/pkg/@warp-ds/tokens/v1/finn-no.css" />
```
magnuh, line 12: Should we use Finn as the base theme here? Or should there be like a `no-brand-theme.css`?

pearofducks: There should be a no-brand-theme used here, but I don't know if design has that done yet?

magnuh: We will go ahead with finn for now and change it later. 👍

Skadefryd: Prob easier to use the finn theme for now, since we know what that is supposed to look like.

AnnaRybkina (approved):

BalbinaK:
[warpify react repo](https://trello.com/c/SwTwYlAH/136-warpify-react-repo)

PR #6, opened 2023-03-23
chore: initial change from fabric core to warp and setting up things for warp drive

magnuh:
Co-authored by @BalbinaK

### DIFF
--- a/.storybook/main.cjs
+++ b/.storybook/main.cjs
@@ -3,7 +3,7 @@ module.exports = {
         builder: 'webpack5',
     },
     stories: ['../packages/**/*.stories.[tj]sx'],
-    addons: ['@storybook/addon-actions/register', '@storybook/addon-postcss'],
+    addons: ['@storybook/addon-actions/register',/* '@storybook/addon-postcss'*/],
     framework: '@storybook/react',
     webpackFinal: async (config) => {
         config.resolve.alias['@warp-ds/core/attention'] = require.resolve(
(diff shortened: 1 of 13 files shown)


vite.config.js
```diff
@@ -13,6 +15,9 @@ export default function config({ mode }) {
       mdx.default({
         remarkPlugins: [slug, autoLinkHeadings, jsxExample, propTable],
       }),
+      uno({
+        presets: [presetWarp({ usePixels: true, usePreflight: true })]
```
pearofducks, line 19: Just curious why this is using pixels? Does storybook break with the REM root change?

.storybook/main.cjs
```diff
@@ -0,0 +1,34 @@
+module.exports = {
+    core: {
+        builder: 'webpack5',
```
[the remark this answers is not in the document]
magnuh: Yup it will. 😅  ...that's the way it was done in Fabric. But it was a lot more problems with Storybook (as I mentioned earlier)... this was just an attempt at getting it running at lowest cost. 😅  We decided to save it for later and see what we do with it...

vite.config.js
magnuh: It kind of broke, if I remember it correctly. This was the easy way forward, I guess, as Fabric/react used pixels? @BalbinaK?
But we should change to using rem from now on?

pearofducks: We definitely should use REM if possible, and if not possible using Storybook that might put a stronger emphasis on moving to Ladle (if it would support REM better).

This environment is where we're developing and ensuring components are behaving as they should - and part of that is making sure accessibility features like resizes work properly.

Using pixels will break that, and make it rather difficult to ensure React components behave as they will in real world scenarios. Unless ofc y'all have some workaround in mind to ensure this behavior works.

Skadefryd (approved):

vite.config.js
Skadefryd: Yea I agree, but not a shostopper for this pr, is it ? :)

magnuh: We'll merge this and then have another look at changing it to rem. 👍

BalbinaK:
[warpify react repo](https://trello.com/c/SwTwYlAH/136-warpify-react-repo)

PR #8, opened 2023-03-24
fix(alert): don't use dynamic classes

BalbinaK:
## Description
Fixes an issue mentioned in https://github.com/warp-ds/react/pull/5
> The automatic class factory doesn't work when trying to dynamically create the classes like using a prop in react for setting the type i-border-$color-alert-${type}-subtle-border...but it does work once I have used the class name hard coded before doing it dynamically, like i-border-$color-alert-success-subtle-border. Guessing that the css compiler has already run when the class gets set in react... 🤷

## Notes
The `alert` object will eventually be moved to https://github.com/warp-ds/component-classes but we might want to wait with that until border-left-color is applicable through a utility class so we can add it to the strings.

In a separate PR I will rename `critical` and `success` variants back to `negative` and `positive` to make it easier to revert those changes the future if we want those restored.

Screenshot 2023-03-24 at 14 23 46

### DIFF
--- a/packages/alert/src/component.tsx
+++ b/packages/alert/src/component.tsx
@@ -11,18 +11,21 @@ export function Alert({
   ...props
 }: PropsWithChildren<AlertProps>) {
 
+  const variantClasses = alert[type]
+  const iconVariantClasses = alert[`${type}Icon`]
+
   return (
     <ExpandTransition show={show}>
       <div
         role={role}
         className={classNames(
           props.className,
-          `flex p-16 border border-l-4 rounded-4 i-border-$color-alert-${type}-subtle-border i-bg-$color-alert-${type}-background i-text-$color-alert-${type}-text`,
+         `${alert.alert} ${variantClasses}`,
         )}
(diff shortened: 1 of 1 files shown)


packages/alert/src/component.tsx
```diff
@@ -11,18 +11,21 @@ export function Alert({
   ...props
 }: PropsWithChildren<AlertProps>) {
 
+  const variantClasses = alert[type]
+  const iconVariantClasses = alert[`${type}Icon`]
```
magnuh, line 15: Not sure how react works, but if you change the prop from where it is used from let's say `negative` to `positive`... will this then be re-evaluated so that the classes will be updated to reflect the change? 🤔 🤷

BalbinaK: I removed any mentions of warning variant alerts in the site, and then updated the variant to "warning" on button click. Looks like updating the classes to use `warning`-variant tokens worked on runtime!
variant-updating

magnuh (approved):

PR #9, opened 2023-03-24
revert(alert): use negative,positive variants instead of critical,success

BalbinaK:
Originally the naming change to `critical` and `success` was made to try and align things more correctly as a system but we decided to scope breaking changes down in Warp 1.0.

These changes require respective tokens to be renamed and published in `@warp-ds/tokens` so until that's done, this will be a draft PR.


BalbinaK:
[MOB: Apply tokens to Alert components](https://trello.com/c/QKIvm0Qr/118-mob-apply-tokens-to-alert-components)

PR #15, opened 2023-04-05
fix: use updated component-classes in alert component

BalbinaK:
## Changes:
- bump `@warp-ds/component-classes` to `1.0.0-alpha.15` to apply updated styles to the alert icon ([changelog](https://github.com/warp-ds/component-classes/blob/alpha/CHANGELOG.md#100-alpha15-2023-04-05))
- fix class for controlling bottom margin of last child (`last-child:` was renamed to `last:` in @warp-ds/uno)
- don't use pixels in `presetWarp` as we want to default REM values when displaying Warp components. It will cause components to look a bit off compared to fabric components, but that's until we handle typography

Screenshot 2023-04-05 at 10 41 34

### DIFF
--- a/.storybook/main.cjs
+++ b/.storybook/main.cjs
@@ -28,7 +28,7 @@ module.exports = {
         const { classes } = (await import('@warp-ds/component-classes/classes'));
 
         config.plugins.push(uno.default({
-            presets: [presetWarp({ usePixels: true, usePreflight: true })],
+            presets: [presetWarp({ usePreflight: true })],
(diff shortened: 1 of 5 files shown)


BalbinaK:
[Change from pixels to REM values for Warp React](https://trello.com/c/aWM4yUAM/162-change-from-pixels-to-rem-values-for-warp-react)

AnnaRybkina (approved):
Awesome 🎆

PR #16, opened 2023-04-05
fix: apply new tokens to textfield and textarea

AnnaRybkina:
Styles were migrated from https://github.com/fabric-ds/css/blob/49b07c05686ef132b203026e941bf04dbab6fc7a/src/components/form-elements.css

For testing in docs:
`http://localhost:3000/textarea`
`http://localhost:3000/textfield`
For testing in Storybook:
`http://localhost:9003/?path=/story/forms-textfield--standard` - check all types
`http://localhost:9003/?path=/story/forms-textarea--standard` - check all types

Design here: https://www.figma.com/file/8P1JQsd82b93gQ6K3igO2p/Warp---Component-library?node-id=393-38455&t=2Rt4QOodGYjj5KNF-0
Old docs: https://react.fabric-ds.io/textfield

### DIFF
--- a/packages/textarea/src/component.tsx
+++ b/packages/textarea/src/component.tsx
@@ -39,27 +39,41 @@ export const TextArea = forwardRef<HTMLTextAreaElement, TextAreaProps>(
       minimumRows,
     });
 
+    const input = {
+      default: 'block text-16 mb-0 leading-22 i-text-$color-input-text-filled i-bg-$color-input-background i-border-$color-input-border hover:i-border-$color-input-border-hover active:i-border-$color-input-border-active rounded-4 py-12 px-8 block border-1 w-full focusable',
+      optional: 'pl-8 font-normal text-14 i-text-$color-label-optional-text',
+      disabled: 'i-bg-$color-input-background-disabled i-border-$color-input-border-disabled i-text-$color-input-text-disabled pointer-events-none',
+      invalid: 'focusable i-border-$color-input-border-error i-text-$color-input-text-error',
+      readOnly: 'pl-0 bg-transparent border-0 pointer-events-none i-text-$color-input-text-read-only',
+      label: 'antialiased block relative text-14 font-bold pb-4 cursor-pointer',
+      labelValid: 'i-text-$color-label-text',
+      labelInvalid: 'i-text-$color-helptext-error-text',
+      helpText: 'text-12 mt-4 block',
+      helpTextValid: 'i-text-$color-helptext-text',
+      helpTextInvalid: 'i-text-$color-helptext-error-text'
+    }
+
     return (
       <div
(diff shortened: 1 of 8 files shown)


packages/textarea/src/component.tsx
```diff
@@ -41,25 +42,25 @@ export const TextArea = forwardRef<HTMLTextAreaElement, TextAreaProps>(
...
             {optional && (
-              <span className="pl-8 font-normal text-14 text-gray-500">
+              <span className={input.optional}>
                 (valgfritt)
               </span>
             )}
           </label>
         )}
         <textarea
+          className={classNames({
```
BalbinaK, line 58: In the previous version `className` prop was also included. Was it intentionally removed?

AnnaRybkina: Which version 😊 I can add `className` prop. Do we need it? Like that you mean? 👇 
```suggestion
          className={classNames(props.className,{
```

BalbinaK: Yes, that's what was there before as I see in your changed files (by previous version I meant `master` version). Not sure if that's still needed but I thought we should be careful about removing stuff like that in case we break something for users that need to migrate from Fabric to Warp. Another thing is that in `master` the `className` prop was applied to the wrapping `div`, not the `textarea` element. Whether that makes things behave differently - I'm again not sure.

AnnaRybkina: Ah do you meant here https://github.com/warp-ds/react/pull/16/files#diff-e6dfac355d15645c03e0c5dea277d37e1d18bb5460789f49189f9c8ab4b130fdL44?
I've "refurnitured" a bit here as for example this one https://github.com/fabric-ds/css/blob/49b07c05686ef132b203026e941bf04dbab6fc7a/src/components/form-elements.css#L42 has `& input` inside the class and that is not what we have now, right? As we want to have classes for specific elements. I've tested and the look and feel is preserved but maybe I'm missing something

BalbinaK: Moving the invalid/disabled/readonly classes down to respective elements makes sense, yes. What I'm not sure about is whether we should keep `className` prop on the textarea element or rather the wrapping div like it was before. So something like this:
```
<div
   className={className}
   style={style}
>
```

Does that make sense?

AnnaRybkina: Ok, I see. Let's keep it on the parent. Good catch 👍

AnnaRybkina: Fixed https://github.com/warp-ds/react/pull/16/files#diff-e6dfac355d15645c03e0c5dea277d37e1d18bb5460789f49189f9c8ab4b130fdR45

PR #21, opened 2023-04-19
fix: add component classes to select

AnnaRybkina:
Migrated styling from https://github.com/fabric-ds/css/blob/49b07c05686ef132b203026e941bf04dbab6fc7a/src/components/form-elements.css

- Removed `labelValid` as it only was a color and is now a part of the default label styling

Variant|Screenshot
---|---
Standard | image
Hint | image
Invalid | image
Disabled | image
No label | image
Optional | image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "@warp-ds/react",
   "version": "1.0.0-alpha.14",
+  "type": "module",
   "repository": "git@github.com:warp-ds/react.git",
   "license": "Apache-2.0",
   "exports": {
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "@warp-ds/react",
   "version": "1.0.0-alpha.14",
-  "type": "module",
```
BalbinaK, line 4: Oopsy? 😄

BalbinaK (approved):

package.json
AnnaRybkina: Damn it 😶‍🌫️

AnnaRybkina: Fixed now

PR #23, opened 2023-04-21
fix: add component classes to Breadcrumbs

BalbinaK:
## Description
- update the breadcrumbs component to using [component-classes](https://github.com/warp-ds/component-classes/blob/alpha/index.js#L310-L315)
- color tokens work
- remove `space-x-reverse` class from the wrapper - this class was applied before to fix a [bug](https://github.com/fabric-ds/react/issues/57) that now is gone when wrapper classes are applied to a different element (like in [Warp Vue](https://github.com/warp-ds/vue/blob/alpha/components/breadcrumbs/w-breadcrumbs.vue#L4)).

## Additionally
- render `string`-typed children as styled `span` elements with `aria-current="page"` on last child
- 
Screenshot 2023-04-21 at 10 58 51


packages/breadcrumbs/docs/Breadcrumbs.mdx
```diff
@@ -16,9 +16,9 @@ import { Breadcrumbs } from '@warp-ds/react';
 <Breadcrumbs>
   <a href="#/url/1">Eiendom</a>
   <a href="#/url/2">Bolig til salgs</a>
-  <a href="#/url/3" aria-current="page">
+  <span href="#/url/3" aria-current="page">
```
magnuh, line 19: Should this still have the href attribute when converted to a span?

BalbinaK: Ah, good catch! Fixed in [21deafd](https://github.com/warp-ds/react/pull/23/commits/21deafddb91497c85355dbd407796b21d80b78d6)

PR #28, opened 2023-05-04
fix: Add Warp component-classes to the Pill component

magnuh:


BalbinaK (commented):
We need to uncomment the export of pill in `index.ts`, too 😊 Other than that LGTM!

CarolineGard (commented):
Nice!! 
There are some height differences in the pill when adding icon,
Screenshot 2023-05-05 at 08 41 18
Screenshot 2023-05-05 at 08 40 51

magnuh:
> We need to uncomment the export of pill in `index.ts`, too 😊 Other than that LGTM!

Ah perfect! Thanks! 🙂 🙏

magnuh:
> Nice!! There are some height differences in the pill when adding icon. There are different paddings for the icon and the string.. it's a bit strange that the height depends on the padding here for the button and div independently, but this solution came from fabric? Screenshot 2023-05-05 at 08 41 18 Screenshot 2023-05-05 at 08 40 51

This is because of no typography. 😄  ...will work when the text-12 class is added.

BalbinaK (approved):

PR #29, opened 2023-05-05
chore: fix broken docs after vite upgrade

AnnaRybkina:
Both docs and Storybook should work

Test to run:
`pnpm docs:dev` - should work
`pnpm dev` - should also work


BalbinaK (approved):
LGTM!

PR #31, opened 2023-05-08
fix: radio buttons classes

AnnaRybkina:
The last part of the Toggle component, radio buttons. Buttongroup tokens are used. Styling was migrated from https://github.com/fabric-ds/css/blob/49b07c05686ef132b203026e941bf04dbab6fc7a/src/components/form-elements.css#L201

image

To test locally: http://localhost:9003/?path=/story/forms-toggle-radiobuttons--multiple-options

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,7 +106,7 @@
   },
   "dependencies": {
     "@chbphone55/classnames": "^2.0.0",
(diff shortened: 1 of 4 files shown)


BalbinaK (approved):
Works perfectly! Great job! 🚀

PR #32, opened 2023-05-11
feature(slider): Add slider

CarolineGard:
This PR includes:
- Warpify slider component
- Fix for making it possible to change the slider from outside the component, based on: https://github.com/fabric-ds/react/pull/187 

Design question: Should we remove the `box-shadow` from the thumb when `slider` is disabled? If yes, that will be fixed in `component-classes`

### DIFF
--- a/packages/slider/stories/Slider.stories.tsx
+++ b/packages/slider/stories/Slider.stories.tsx
@@ -17,6 +17,7 @@ export const Regular = () => {
         max={10_000_000}
         step={1000}
       />
+      <button onClick={() => setValue(2_500_000)}>Reset</button>
     </div>
   );
 };
--- a/packages/slider/stories/Slider.stories.tsx
+++ b/packages/slider/stories/Slider.stories.tsx
@@ -27,7 +27,7 @@ export const Disabled = () => {
     <div>
       <output>{value}</output>
       <Slider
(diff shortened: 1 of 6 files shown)


AnnaRybkina:
Looks good 🎆  

One thing I noticed: I can't see any focus indication on an enabled slider but see it on the disabled. If it is `focusable` that needs to be applied or not I would recommend to add a TODO somewhere to fix focus state so we don't forget to fix it before we say the component is 100% done. `focusable` is working but not as simple as we would like.

packages/slider/stories/Slider.stories.tsx
```diff
@@ -27,7 +27,7 @@ export const Disabled = () => {
     <div>
       <output>{value}</output>
       <Slider
-        onChange={(value) => setValue(value)}
+        onChange={(val) => setValue(val)}
```
AnnaRybkina, line 30: I wouldn't mind adding https://github.com/fabric-ds/react/pull/187/files#diff-0be3b4850e403dc499c4a02e50cdde7ceb13fc901a97da54b82150cddfd867c1R20 in this story to be able to easily test the fix.

CarolineGard: yes nice!

PR #34, opened 2023-05-15
feat(tabs): warpify tabs

CarolineGard:
This PR includes:
- Warpify `Tabs`
- Remove `contained` as an option for tabs: This decision was made by design while no/very few teams are currently using it and it's not part of the current and future design. 
- Style changes according to new design: e.g. grey border on hover, no underline on hover
- Temporary solution for handling the number of grid columns: adding classes dynamically causes issues as it prevents the CSS class to load properly from the drive


packages/tabs/src/component-tab.tsx
```diff
@@ -1,32 +1,20 @@
...
 
-const setup = ({
-  className,
-  isActive,
-  setActive,
-  contained,
-  ...rest
-}: any) => ({
-  tab: cn({
+const setup = ({ className, isActive, setActive, ...rest }: any) => ({
```
AnnaRybkina, line 6: I would really prefer to either fix eslint config for the repo or switch off local formatting.

packages/tabs/src/component-tabs.tsx
```diff
@@ -6,31 +6,31 @@ import React, {
...
 const setup = (
   { className, contained, children, onClick, active, ...rest }: any,
   tabsRef,
-  wunderbarRef,
+  wunderbarRef
 ) => ({
   nav: cn({
     [className]: !!className,
-    [contained ? c.wrapperContained : c.wrapperUnderlined]: true,
+    [contained ? ccTabs.wrapperContained : ccTabs.wrapperUnderlined]: true,
```
AnnaRybkina, line 20: I guess `contained` doesn't matter here anymore?

```suggestion
    [ccTabs.wrapperUnderlined]: true,
```

packages/tabs/src/component-tab.tsx
```diff
@@ -1,32 +1,20 @@
...
+  tab: cn([ccTab.tab], {
```
AnnaRybkina, line 7: Minor, I think you can even skip the array here
```suggestion
  tab: cn(ccTab.tab, {
```

packages/tabs/src/component-tab.tsx
```diff
@@ -1,32 +1,20 @@
...
-    [c.tabContainedActive]: contained && isActive,
+    [ccTab.tabActive]: isActive,
   }),
-  icon: cn({
-    [c.icon]: true,
-    [c.iconUnderlined]: !contained,
-    [isActive ? c.iconUnderlinedActive : c.iconUnderlinedInactive]: !contained,
+  icon: cn([ccTab.icon], [ccTab.iconUnderlined], {
+    [isActive ? ccTab.iconUnderlinedActive : ccTab.iconUnderlinedInactive]:
+      true,
```
AnnaRybkina, line 13: Minor, this could be refactored as 👇 Wdyt?
```suggestion
  icon: cn(ccTab.icon, ccTab.iconUnderlined, isActive ? ccTab.iconUnderlinedActive : ccTab.iconUnderlinedInactive)
```

AnnaRybkina (approved):
Great work 💪 I left some minor comments but leave those are up to you to decide if the suggestions make sense 😁

PR #37, opened 2023-05-25
feat(attention): warpify attention

CarolineGard:
Warpify attention component. 

Uses the same solution as in fabric with `filter: drop-shadow(...)` for `popover` to have a shadow affecting both the box and the arrow. 

Screenshot 2023-05-25 at 10 41 44


packages/attention/src/component.tsx
```diff
@@ -91,47 +100,52 @@ export function Attention(props: AttentionProps) {
...
-          [c.callout]: props.callout,
-          [c.popover]: props.popover,
-        })}
-      >
+      <div className={wrapperClasses}>
         {!props.noArrow && (
           <Arrow {...props} ref={arrowRef} direction={placement} />
         )}
-        <div className="last-child:mb-0">{props.children}</div>
+        <div className="last:mb-0">{props.children}</div>
```
AnnaRybkina, line 107: Please add this one to the component-classes

packages/attention/src/component.tsx
```diff
@@ -91,47 +100,52 @@ export function Attention(props: AttentionProps) {
...
+      className={arrowClasses}
       style={{
         // TW doesn't let us specify exactly one corner, only whole sides
         borderTopLeftRadius: '4px',
         zIndex: 1,
         // border alignment is off by a fraction of a pixel, this fixes it
-        [`margin${
-          arrowDirection.charAt(0).toUpperCase() + arrowDirection.slice(1)
-        }`]: '-0.5px',
+        [`margin${arrowDirection.charAt(0).toUpperCase() + arrowDirection.slice(1)}`]: '-0.5px',
```
AnnaRybkina, line 148: Can `arrowDirectionClassname ` be used here? The logic looks the same but I haven't checked more specifically

packages/attention/src/component.tsx
```diff
@@ -91,47 +100,52 @@ export function Attention(props: AttentionProps) {
...
+  const arrowType = () => {
+    if (tooltip) return 'arrowTooltip';
+    else if (callout) return 'arrowCallout';
+    else if (popover) return 'arrowPopover';
+    return '';
+  }
+
+  const arrowClasses = classNames(
+    ccAttention.arrowBase,
+    ccAttention[arrowDirectionClassname(arrowDirection)],
```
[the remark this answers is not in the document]
AnnaRybkina: Ah nvm, I saw now when I tested that it is all good 👍

packages/attention/src/component.tsx
```diff
@@ -91,47 +100,52 @@ export function Attention(props: AttentionProps) {
...
+    {
+      [ccAttention[arrowType()]]: tooltip || callout || popover
```
AnnaRybkina, line 134: Imho it was much more readable before when there were no need for a function to jump into
``` 
   [ccAttention.tooltip]: props.tooltip,
   [ccAttention.callout]: props.callout,
   [ccAttention.popover]: props.popover,
```

If the refactor is necessary I would maybe suggest to add an enum or something and skip the function

AnnaRybkina (approved):

PR #38, opened 2023-05-25
feat: add component classes to the combobox

AnnaRybkina:
https://react.fabric-ds.io/combobox

image


CarolineGard (approved):
looks good and works well

PR #39, opened 2023-05-29
feat: Warpified Switch component

magnuh:


docs/src/Main.tsx
```diff
@@ -86,6 +86,10 @@ const App = () => {
             <Steps />
           </Route>
 
+          <Route path="/switch">
+            <Switch />
+          </Route>
+
```
CarolineGard, line 92: Maybe remove the switch route from below that is commented out, line 122?🙂

packages/switch/src/component.tsx
```diff
@@ -24,18 +24,15 @@ export function Switch({
...
-        className={classNames([c.label, switchFocus])}
+        className={classNames([ccSwitch.label, switchFocus])}
         {...attrs}
       >
         <span
-          className={classNames([c.switchTrack, 'top-0', 'left-0'], {
-            [c.switchTrackSelected]: value,
-            [c.switchTrackUnselected]: !value,
-          })}
+          className={classNames([ccSwitch.track, 'top-0', 'left-0', value ? ccSwitch.trackActive : ccSwitch.trackInactive])}
```
CarolineGard, line 31: is there a reason to not add `top-0` and `left-0` to ccSwitch.track classes?

CarolineGard (approved):
Works well! 
just comments on some small tweaks 🙌

packages/switch/src/component.tsx
magnuh: Yup, because the React versions of the Switch component is made up of a button and span elements, unlike the Vue equivalent which uses an input and divs. So this is the way it was in Fabric as well...

docs/src/Main.tsx
magnuh: Ah of course. 😄 👍

magnuh: done

packages/switch/src/component.tsx
CarolineGard: aha okay cool, let's do it like this then!

PR #49, opened 2023-06-15
fix(components): update component classes

CarolineGard:
Affected components:
attention, tabs, expandable, steps


packages/expandable/src/component.tsx
```diff
@@ -56,7 +56,7 @@ export function Expandable(props: ExpandableProps) {
           })}
           onClick={() => toggleExpandable(stateExpanded)}
         >
-          <div className="flex justify-between align-center">
+          <div className={ccExpandable.title}>
             {typeof title === 'string' ? (
               <span className="h4">{title}</span>
```
BalbinaK, line 61: Should we move this class, too?

PR #50, opened 2023-06-16
fix(packages): use className and style in TextField and Select

BalbinaK:
Fixes an issue described [in this Slack thread](https://sch-chat.slack.com/archives/C04P0GYTHPV/p1686906779445769) where className set on `Select` or `TextField` would override the whole `className` prop of the `input` element.

Also added support for styling the whole component with `style` prop, as it seems to have been the strategy for other form components (Select, Textarea)

### DIFF
--- a/packages/select/src/component.tsx
+++ b/packages/select/src/component.tsx
@@ -46,10 +46,7 @@ const setup = (props) => {
             }
           : null,
     },
-    wrapperClasses: classNames({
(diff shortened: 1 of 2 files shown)


BalbinaK:
[TextField - className prop overrides our component classes](https://trello.com/c/haaSakIx/307-textfield-classname-prop-overrides-our-component-classes)

AnnaRybkina (approved):

PR #51, opened 2023-06-20
fix: remove background from checkbox/toggle

AnnaRybkina:
If the checkbox is used on a container with a background that is different than white, the whole input gets a white background with i-bg-$color-checkbox-background. Not needed anymore, gone in Vue. Forgot to remove it from here


magnuh (approved):

AnnaRybkina:
Before|After
---|---
image|image

PR #53, opened 2023-06-22
fix: use correct padding for small buttons

siljec:
**Issue**: different padding for small buttons for primary, negative and utility. The height should always be 32px (y padding 8 px) for small buttons including border. Might be better to use same padding for all small buttons and use `box-sizing: border-box` though 🤔 

**Solution:** use correct classes for small buttons.

```
buttonSmall: 'px-16 py-6 text-xs leading-xs', // .button--small
buttonSmallOverride: 'py-8', // .button--small.button--primary, .button--small.button--destructive, .button--small.button--destructive-flat, .button--small.button--order, .button--small.button--quiet
buttonSmallUtility: 'py-7 px-15', // .button--small.button--secondary
```
**Before (Fabric vs Warp):**
image

**Now (react storybook):**
image

### DIFF
--- a/packages/button/src/component.tsx
+++ b/packages/button/src/component.tsx
@@ -39,7 +39,9 @@ export const Button = forwardRef<
     [ccButton.buttonUtilityFlat]: utility && quiet,
     // others
     [ccButton.buttonSmall]: small,
+    [ccButton.buttonSmallOverride]: small && !(secondary || utility),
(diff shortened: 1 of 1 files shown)


AnnaRybkina (approved):
Nice catch and fix 👍 🚢

CarolineGard (approved):
Nice 🙌

PR #55, opened 2023-06-28
fix: use css packages instead of component-classes

AnnaRybkina:
Fixes [Warp-185](https://nmp-jira.atlassian.net/browse/WARP-185?atlOrigin=eyJpIjoiNTI4YmYyMzExYzJmNGQ0ZDkyYzg0YmVlODUzYTMzZWMiLCJwIjoiaiJ9)

In this PR we are supporting the new semantic classes that was a breaking change in the @warp-ds/css@1.0.0-alpha.33.

Changes include:

- bump @warp-ds/uno to -alpha.60 & @warp-ds/css to -alpha.33
- import component classes from @warp-ds/css/component-classes
- use new semantic color tokens from @warp-ds/css/1.0.0-alpha.33
- use new resets in uno config


BalbinaK (approved):
🚀

AnnaRybkina:
Found some typing issues when doing the local build. Fixing those now.
image

AnnaRybkina:
I've updated the types and bumped deps again. @BalbinaK Do you want to take one more look before I merge? 🙌

package.json
```diff
@@ -104,8 +103,9 @@
   },
   "dependencies": {
     "@chbphone55/classnames": "^2.0.0",
-    "@warp-ds/component-classes": "^1.0.0-alpha.108",
+    "@warp-ds/css": "^1.0.0-alpha.15",
```
BalbinaK, line 106: Should this also be at 22 like the eik imports?

vite.config.js
```diff
@@ -6,15 +6,16 @@ import jsxExample from './docs/src/plugins/jsxExample.cjs';
...
 
 
 export default function config() {
   return {
     // base: '/',
     plugins: [
       uno({
-        presets: [presetWarp({ usePreflight: true })],
+        presets: [presetWarp({ skipResets: true })],
         safelist: classes,
```
siljec, line 19: Don't we need this code? 

```
let reset;
async function getReset() {
  if (reset) return reset;
  else {
    reset = (await fetch('https://assets.finn.no/pkg/@warp-ds/css/1.0.0-alpha.33/resets.min.css')).text();
    return reset;
  }
}

...
```
I see that we are including the reset in storybook, but not in the index.html 🤔

siljec: oh, I just looked at vue, I guess we want to move the reset to index.html instead of the `.storybook/preview-body.html` file?

siljec: Or is it not necessary to add it, as we expect the teams to add it?  🤔

BalbinaK: `index.html` is only used for an old documentation site (the one you see in https://react.fabric-ds.io/), which we haven't yet removed from this repo. It was mentioned at some point that we might want to remove it but we didn't prioritize it. Until we do that, just for the sake of consistency, I added the resets there, too 😁

`.storybook/preview-body.html` should actually be renamed to `preview-head.html` as its contents belong more in the head of the storybook's component page ([storybook docs](https://storybook.js.org/docs/react/configure/story-rendering#adding-to-head)). If we remove the styles from there completely, we wouldn't have access to the stylesheets needed to load stories.

siljec (approved):

vite.config.js
BalbinaK: Renamed file to `preview-head.html` in [127eb77](https://github.com/warp-ds/react/pull/55/commits/127eb7767dba7b8aa9df9677fdaaeccc35ef9add) and tested that all still looks correct in storybook.

PR #58, opened 2023-07-07
chore: update typescript, eslint and remove microbundle

AnnaRybkina:
Before|After
---|---
image|image

- [x] Remove [microbundle](https://github.com/developit/microbundle) that has most of the deprecated deps as peer deps. I was not sure about this one but I really couldn't find any usages of this package. As far as I. understood you need to add `microbundle` to the config but I couldn't see any of that in here. Please correct me if I'm wrong.
- [x] Upgrade eslint packages to the ones without deprecated deps
- [x] Upgrade typescript,Iäve upgraded, retested and rebuilt and I couldn't see any errors after the upgrade. Please retest/rebuild and check if it is correct

I've tested to use the components in my test app and everything still seems to work as expected.

TODO: we have one more deprecated peer dep and it will be gone after the removal of the docs. I decided to remove those in a separate PR otherwise the diff will be too big


BalbinaK:
> Remove [microbundle](https://github.com/developit/microbundle) that has most of the deprecated deps as peer deps. I was not sure about this one but I really couldn't find any usages of this package.

`microbundle` was replaced with `tsc` at [this point](https://github.com/fabric-ds/react/commit/d653c454ebde67dd2dbae24aeb37d8f421d47b8b#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519) in Fabric, so yes - it should be removed! Good catch!

siljec (approved):
Nice! Looks good to me 👌

PR #59, opened 2023-07-07
fix: stable FINN prod alpha-release

BalbinaK:
## Background
With this setup @warp-ds/react will be compatible with [tokens](https://assets.finn.no/pkg/@warp-ds/tokens/v1/finn-no.css) currently available in FINN production. After merging this PR and so releasing a new alpha version, we will update [Warp Tech Docs](https://warp-ds.github.io/tech-docs/getting-started/developers/) and suggest the use of this particular version when relying on the above mentioned tokens.
 
## Changes
- bump `@warp-ds/uno` to `1.0.0.-alpha.49`
- bump `@warp-ds/component-classes` to `1.0.0.-alpha.115`

Also:
- Fix missing transparent background on buttons in Clickable, Tabs and Button of type "link"
- import classes for Clickable from component-classes

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
   "dependencies": {
     "@chbphone55/classnames": "^2.0.0",
     "@warp-ds/uno": "^1.0.0-alpha.49",
-    "@warp-ds/component-classes": "^1.0.0-alpha.108",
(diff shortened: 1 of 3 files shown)


AnnaRybkina (approved):
Haven't tested but code looks good. Let's 🚢 after the bump

PR #63, opened 2023-08-01
feat: add internationalization (i18n)

siljec:
[WARP-199](https://nmp-jira.atlassian.net/browse/WARP-199): Add internationalization by using [Lingui](https://lingui.dev/introduction). Currently translations are in the code, but it should be moved to a Translation Management System eventually.

How it works:
* In [lingui.config.ts](https://lingui.dev/ref/conf): Define 1) supported locale 2) files to support internationalization for, and 3) specify namespace for exporting compiled messages.
* Add supported locale in `i18n.ts`
* Run `activateI18n` method in the files where you need translation to load the correct translated messages.
* Use [i18n._()-core](https://lingui.dev/ref/core#i18n._) method for translating messages.
* [lingui extract](https://lingui.dev/ref/cli#extract) command extracts messages from code and generates the "po-messages". (Need to run this command when adding new strings to be translated) The "po-messages" need to be translated.
* [lingui compile](https://lingui.dev/ref/cli#compile) command compiles the"po-messages" and generates minified files. Because we have `compileNamespace: 'es'` in `lingui.config.ts` it outputs `.mjs` ((ES6 modules)-javascript) files. (Need to run this command when the po files are updated with correct translation)
* Add i18n workflow - to make sure all strings are defined for all languages in the `.po`-files. However, it's not checking if we remembered to run `lingui compile` and check in the compiled `.mjs`-files...

Other notes:
* Use lang `nb` in storybook
* Looked at the example draft PR [here](https://github.com/fabric-ds/react/pull/184/files) 
* Translations can be found [here](https://docs.google.com/spreadsheets/d/1Kizv8fwRZLDk0N8VkcyPZ7Q22tHTUovLFqLsdYhWFlg/edit#gid=0)


packages/utils/src/useI18n.ts
```diff
@@ -0,0 +1,40 @@
+import { useEffect, useState } from 'react';
+import { dynamicActivateI18n } from './dynamicActivatei18n';
+
+function detectLocale() {
+  const supportedLocales = ['en', 'nb'];
+  const defaultLocale = 'en';
+  try {
+    const htmlLocale = document.documentElement.lang;
```
siljec, line 8: This might not work server side. Need to look into how to get the correct language here.

packages/utils/src/useI18n.ts
```diff
@@ -0,0 +1,50 @@
+import { useEffect, useState } from 'react';
+import { dynamicActivateI18n } from './dynamicActivatei18n';
+
+function detectLocale() {
+  const supportedLocales = ['en', 'nb', 'fi'];
```
BalbinaK, line 5: I see we've defined those in `lingui.config.js`, too. Would it make sense to define it once and import it e.g. here?

BalbinaK (approved):
🚀

PR #64, opened 2023-08-09
fix(select): fix select chevron

siljec:
[WARP-210](https://nmp-jira.atlassian.net/browse/WARP-210) - Previously, nothing happened on hover or onClick on the chevron. @warp-ds/css component-classes has updated select.chevron to include "pointer-events-none" and "cursor-pointer" which solves the issue. 

Also, update brand css and resets to v1, and add fonts stylesheet

### DIFF
--- a/index.html
+++ b/index.html
@@ -9,9 +9,9 @@
       href="https://www.finn.no/favicon.ico"
       type="image/x-icon"
     />
-    <link rel="stylesheet" href="https://assets.finn.no/pkg/@warp-ds/css/v1/tokens/finn-no.css" />
-    <link rel="stylesheet" href="https://assets.finn.no/pkg/@warp-ds/css/v1/resets.css" />
     <link rel="stylesheet" href="https://assets.finn.no/pkg/@warp-ds/fonts/v1/finn-no.css" />
+    <link rel="stylesheet" href="https://assets.finn.no/pkg/@warp-ds/css/1.0.0-alpha.37/tokens/finn-no.css" />
+    <link rel="stylesheet" href="https://assets.finn.no/pkg/@warp-ds/css/1.0.0-alpha.37/resets.css" />
   </head>
   <body>
     <div class="m-16">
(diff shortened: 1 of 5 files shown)


BalbinaK (commented):
Nice! I think similar eik urls changes should be added to `.storybook/preview-head.html` for storybook to work :)

index.html
```diff
@@ -9,8 +9,9 @@
       href="https://www.finn.no/favicon.ico"
       type="image/x-icon"
     />
-    <link rel="stylesheet" href="https://assets.finn.no/pkg/@warp-ds/css/1.0.0-alpha.33/tokens/finn-no.css" />
-    <link rel="stylesheet" href="https://assets.finn.no/pkg/@warp-ds/css/1.0.0-alpha.33/resets.min.css" />
+    <link rel="stylesheet" href="https://assets.finn.no/pkg/@warp-ds/css/v1/tokens/finn-no.css" />
```
BalbinaK, line 12: If we use tokens from v1 (1.0.0-alpha.34) and component-classes from `1.0.0-alpha.37`, changes from `1.0.0-alpha.36` won't be applied due to missing tokens. It's just for internal testing of this repo but maybe until we release 1.0.0 and update the aliased version of @warp-ds/css we should continue referencing specific versions of eik urls. Wdyt?

siljec: ah, my bad! Will fix!

index.html
```diff
@@ -9,8 +9,9 @@
...
+    <link rel="stylesheet" href="https://assets.finn.no/pkg/@warp-ds/css/v1/resets.css" />
+    <link rel="stylesheet" href="https://assets.finn.no/pkg/@warp-ds/fonts/v1/finn-no.css" />
```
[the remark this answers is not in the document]
siljec: Makes sense! changed order now 😄

BalbinaK (approved):
💪

PR #66, opened 2023-08-24
fix: button classes refactoring

AnnaRybkina:


packages/button/src/component.tsx
```diff
@@ -28,25 +28,80 @@ export const Button = forwardRef<
     ...rest
   } = props;
 
+   const ccButton = {
```
AnnaRybkina, line 31: This one is checked in ONLY for code reviewing and testing purposes. This must be deleted from here before merge and moved to css repo.

siljec:
> Maybe it is better to create classes which holds everything for a button type + variant.

I like that we create one button class that can be reused several places, so that we don't have to repeat e.g. `border-2 font-bold rounded-8 max-w-max focusable justify-center transition-colors ease-in-out` for all types/variants. I'm thinking it can be error prone when me make changes if we duplicate this, and will need to add classnames several "button classes" - might be easy to mess up.

However, it would make it more readable in React/Vue/Elements! So I'm a bit torn here 🤔

packages/button/src/component.tsx
```diff
@@ -28,25 +28,79 @@ export const Button = forwardRef<
...
+      'border-transparent! animate-inprogress i-text-$color-button-loading-text! pointer-events-none i-bg-$color-button-loading-background!', // .button--in-progress, a.button--in-progress:visited
+    linkAsButton: 'inline-block hover:no-underline',
+    a11y: 'sr-only',
+  };
+
+  const buttonSecondaryDefault = secondary && !quiet || !buttonTypes.find(b => !!props[b]);
+  const buttonDestructive = negative && !quiet;
+  const buttonFlat = secondary && quiet;
+  const buttonDestructiveFlat = negative && quiet;
+  const buttonUtilityFlat = utility && quiet;
```
siljec, line 76: Not saying that we need to fix this now, but I don't like that we are using both destructive/negative and flat/quiet, it's a bit confusing 😅 Would be nice to call the button class e.g. `buttonNegative` and `buttonQuiet` to match the props the user is setting.

AnnaRybkina: yes, totally agree. We have discussed it before. We cannot do that now as shortcuts are used with the old names and changes will bring breaking changes. And in the beginning of the project we said, or at least I want to remember that, that no breaking changes were allowed. Maybe I don't remember the details and maybe it is doable but it should for sure be done when possible

PR #67, opened 2023-08-25
fix: use new button classes and improve conditions

AnnaRybkina:
One more try for button refactoring

In this case we add classes specifically for a button with all variants and such.

The code grew a lot. But I hope it is easier to follow and will improve our work when we will debug problems with buttons.

TODO:

- [x] Move everything to the css repo, ONLY kept here for testing purposes in the beginning (line 28-154 will be moved out)
~- [ ] Shortcuts (what it is in comments) will be fixed in the css repo~

PR for the classes and shortcuts: https://github.com/warp-ds/css/pull/44

First attempt [here](https://github.com/warp-ds/react/pull/66) but it was too unsafe but it is for sure less code but what is more important here 🤷 

- [x] THIS PR SHOULD ALSO BE OPENED TO NEXT BRANCH WHEN IT IS DONE

### DIFF
--- a/packages/button/src/component.tsx
+++ b/packages/button/src/component.tsx
@@ -3,14 +3,6 @@ import { button as ccButton } from '@warp-ds/css/component-classes';
 import { classNames } from '@chbphone55/classnames';
 import type { ButtonProps } from './props';
 
-const buttonTypes = [    
-  'primary',
-  'secondary',
-  'negative',
-  'utility',
-  'pill',
-  'link',
-] as const;
 export const Button = forwardRef<
   HTMLButtonElement | HTMLAnchorElement,
   ButtonProps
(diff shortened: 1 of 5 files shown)


packages/button/src/component.tsx
```diff
@@ -28,25 +28,73 @@ export const Button = forwardRef<
     ...rest
   } = props;
 
+  const buttonVariants = {
```
AnnaRybkina, line 31: This is stuff that will be reused. Therefor a separate const

AnnaRybkina:
@siljec @BalbinaK A small POC(only two buttons fixed) with the second suggestion I had in my previous PR for the button refactoring. Wdyt?

packages/button/src/component.tsx
```diff
@@ -28,25 +28,73 @@ export const Button = forwardRef<
...
+    buttonPrimarySmall:
+      'py-8 px-16 text-xs ' + buttonVariants.primaryDefault,
+    buttonPrimarySmallQuiet:
+      'py-8 px-16 text-xs ' + buttonVariants.buttonFlat,
+    buttonPrimaryLoading:
+      'py-12 px-16 text-m leading-[24] ' + buttonVariants.primaryDefault + buttonVariants.buttonInProgress,
+    buttonPrimarySmallLoading:
+      'py-8 px-16 text-xs  ' + buttonVariants.primaryDefault + buttonVariants.buttonInProgress,
+    buttonPrimaryQuietLoading:
+      'py-12 px-16 text-m leading-[24]  ' + buttonVariants.buttonFlat + buttonVariants.buttonInProgress,
```
AnnaRybkina, line 74: This can also be re-used
```suggestion
     buttonVariants.primarySizing + buttonVariants.buttonFlat + buttonVariants.buttonInProgress,
```

packages/button/src/component.tsx
BalbinaK: I think it makes it clearer to separate the reusable groups of classes. Could we not call it `buttonVariants` though? This is what we call the attribute of Button Element component that can have one of these values: "primary|negative|utility|pill...". Might be we align on the props in Vue and React the same way one day and then these classes might be confusing.

Since we can group these classes in any way possible within this file, maybe we could have the following objects with reusable classes:
- `buttonSizes` (keys like: secondaryMedium, secondarySmall, utilityMedium, utilitySmall, medium, small),
- `buttonModifiers` (flat, inProgress, isDisabled) (cannot think of a better name)
- `buttonVariants (primary, secondary, negative, utility, pill, link)
And then link to those. 

Finally, do you see any issues with us dropping the `button` prefix in the final ccButton keys?

AnnaRybkina: I haven’t spent time on the naming as I was time boxing this POC. Just wanted to visualize what the scope would be. 
I like your suggestions. So you think we should proceed with this solution?

packages/button/src/component.tsx
```diff
@@ -60,7 +223,7 @@ export const Button = forwardRef<
     /*i18n*/ {
       id: 'button.aria.loading',
       message: 'Loading...',
-      comment: 'Screenreader message for buttons that are loading',
+      comment: 'Screenreader message for s that are loading',
```
magnuh, line 226: "s"? 😄

packages/button/stories/Button.stories.tsx
```diff
@@ -18,7 +18,7 @@ export const Example = () => {
         </Button>
         {/* @ts-ignore */}
         <Button className="mr-32" primary loading>
-          Loading
+          Loading test
```
magnuh, line 21: "test"? 🤷😜

AnnaRybkina: Fixed, tnx

siljec:
I'm not a huge fan of the fact that we are adding a lot more code to handle this, but I agree that it's nice to avoid to use important to override "same" classes! But the buttons looks good, and I agree that it will be easier to find issues 💪

pnpm-lock.yaml
```diff
@@ -16,7 +16,7 @@ dependencies:
     version: 1.0.0
   '@warp-ds/css':
     specifier: ^1.0.0
-    version: 1.0.0
+    version: link:../css-1
```
BalbinaK, line 19: oopsy 😁

BalbinaK: Ah, the css PR needs to be out first! Will have a look there.

AnnaRybkina: Classic 🙄 Fixed

BalbinaK (approved):

PR #83, opened 2023-09-06
fix: add check for children in tabs

AnnaRybkina:
Background: https://sch-chat.slack.com/archives/C04P0GYTHPV/p1694002177060589

Do not break when a child is falsy.

image
image

### DIFF
--- a/packages/tabs/src/component-tabs.tsx
+++ b/packages/tabs/src/component-tabs.tsx
@@ -146,9 +146,9 @@ export const Tabs = (props: TabsProps) => {
         onKeyDown={handleKeyDown}
       >
         {Children.map(children, (child: any) => {
-          return cloneElement(child, {
-            setActive: change,
-            isActive: child.props.name === active,
+          return child && cloneElement(child, {
+                setActive: change,
+                isActive: child?.props?.name === active,
           });
         })}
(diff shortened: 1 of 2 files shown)


packages/tabs/src/component-tabs.tsx
```diff
@@ -146,9 +146,9 @@ export const Tabs = (props: TabsProps) => {
         onKeyDown={handleKeyDown}
       >
         {Children.map(children, (child: any) => {
-          return cloneElement(child, {
-            setActive: change,
-            isActive: child.props.name === active,
+          return child && cloneElement(child, {
+                setActive: change,
+                isActive: child?.props?.name === active,
```
Eghizio, line 151: Is it tested for the case when the child is just a regular element like `<div/>` ?
Otherwise it will output such error:
```
React does not recognize the `isActive` prop on a DOM element.
If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `isactive` instead.
If you accidentally passed it from a parent component, remove it from the DOM element.
```

Eghizio (approved):

packages/tabs/src/component-tabs.tsx
AnnaRybkina: Code|Output
---|---
image|image

PR #84, opened 2023-09-08
chore: set up synchronisation with crowdin

j4hr3n:


.github/workflows/crowdin-sync.yml
```diff
@@ -0,0 +1,29 @@
+name: Crowdin Synchronisation
+
+permissions:
+  contents: write
+  pull-requests: write
+
+on:
+  push:
+    branches: [main]
```
siljec, line 9: `next` is our default branch, should we add `next` here too?

j4hr3n: > `next` is our default branch, should we add `next` here too?

@siljec Updated to include `next` now too 👍

siljec (approved):
🎉

PR #85, opened 2023-09-11
feat: Add Badge component

magnuh:
Figma: https://www.figma.com/file/8P1JQsd82b93gQ6K3igO2p/Warp---Components?type=design&node-id=381-42423&mode=design&t=fRLDoRSCmQEMZ7Si-0

This PR uses the changes published through the following PR:s:
https://github.com/warp-ds/drive/pull/156 (@warp-ds/uno v.1.1.0)
https://github.com/warp-ds/css/pull/48 (@warp-ds/css v1.1.0)


packages/badge/src/component.tsx
```diff
@@ -0,0 +1,50 @@
...
+        {
+            [ccBadge.positionBase]: !!position,
+            [ccBadge.positionTL]: position === 'tl',
+            [ccBadge.positionTR]: position === 'tr',
+            [ccBadge.positionBR]: position === 'br',
+            [ccBadge.positionBL]: position === 'bl',
+        },
+        props.className,
+      ),
+    },
```
BalbinaK, line 47: Not sure about this one but I can see Shoelace web component library sets `role="status"` on their badge component ([see here](https://github.com/shoelace-style/shoelace/blob/next/src/components/badge/badge.component.ts#L44C9-L44C22)). Here's some info about role=status from WCAG docs: https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA22
It's not easy for me to see if that applies to our Badge but maybe the someone from the accessibility group could help us here. Wdyt?

magnuh: I guess that is totally up to the consumers what they use it for? 
In Vue they can easily just do `<w-badge role="status">A badge with role</w-badge>`... but it is not as easy in React or Elements? Do we need to add props for this? 🤷 Personally I would think that it should be up to the consumers to add any attribute they want on any component's wrapping element?

magnuh: Badge is also used as a price tag... which I guess wouldn't make sense to put `aria-label="status"` to?

Skadefryd: Ideally we should prob not use this component for pricetag.  And ideally it should probably be role="status" if that is always the case.  Im a bit conflicted about setting Aria roles on stuff as defaults when its not always correct, in fact that creates more issues than it solves in most cases.

Skadefryd (approved):
Looks great :)

PR #89, opened 2023-09-20
fix: Add full width button prop

felicia-haggqvist:
This adds a fullwidth prop to the button component.


BalbinaK (approved):
Lgtm! 
I would maybe update unocss separately next time to avoid it breaking something in our packages. It is still v0 so minor updates could in theory be problematic for us :) It did happen when updating to v0.55 before. It resulted in `@warp-ds/uno` not working correctly, although all worked fine within component repos.

PR #90, opened 2023-09-28
chore: reuse the workflow for crowdin sync

AnnaRybkina:
To avoid the [a bigger amount of code to be added in every repo](https://github.com/warp-ds/vue/pull/65/files#diff-7209abc5864519d00079d00aa93b020b2bc820640edffdc93df1824a2cfc34a7) let's create a reusable workflow and call it to our workflow. We've done something simliar(but a composite action approach) for PR linter to avoid copy pasting the same code everywhere, and it is also easier to maintain in one place

Reusable workflow is created here https://github.com/warp-ds/reusable-workflows

### DIFF
--- a/.github/workflows/crowdin-sync.yml
+++ b/.github/workflows/crowdin-sync.yml
@@ -7,9 +7,8 @@ on:
 jobs:
   synchronize-with-crowdin:
     uses: warp-ds/reusable-workflows/.github/workflows/crowdin-sync.yml@main
-    with:
+    secrets:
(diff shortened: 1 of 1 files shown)


.github/workflows/crowdin-sync.yml
```diff
@@ -1,29 +1,8 @@
 name: Crowdin Synchronisation
 
-permissions:
-  contents: write
-  pull-requests: write
-
 on:
-  push:
-    branches: [next, main]
+  workflow_dispatch:
```
AnnaRybkina, line 4: just for testing for now

magnuh (approved):

PR #95, opened 2023-10-06
refactor: replace inline svgs with warp icons

felicia-haggqvist:
Replace inline-svg:s with the icons from @warp-ds/icons in order to ensure translations for the svg-title will be handled in the same way for all components that use these icons.


package.json
```diff
@@ -92,8 +91,9 @@
     "@chbphone55/classnames": "^2.0.0",
     "@lingui/core": "^4.3.0",
     "@warp-ds/core": "^1.0.0",
-    "@warp-ds/css": "^1.1.0",
+    "@warp-ds/css": "^1.1.1-next.3",
     "@warp-ds/uno": "^1.1.0",
+    "@warp-ds/icons":"1.1.0-next.10",
```
BalbinaK, line 96: If you change that to `1.1.0-next.11` and install 1.1.1 of @warp-ds/css then I'm pretty sure we can merge this PR 🤞

felicia-haggqvist: Updated the versions! 💃

imprashast (approved):

BalbinaK (approved):

PR #118, opened 2023-10-11
chore: remove unused deps

AnnaRybkina:
Couldn't find any usages of these dependencies. I have rebuilt and ran Storybook. All works as before.


imprashast (approved):
Good cleanup 👍

PR #121, opened 2023-10-11
chore: remove unused dependencies

AnnaRybkina:
Couldn't find any usages of these dependencies. I have rebuilt and ran Storybook. All works as before.

What I could see, all of the removed deps have been used in the docs. We have removed the docs


AnnaRybkina:
Something broke when I fixed conflicts in https://github.com/warp-ds/react/pull/118 so I create this one and closing the first one

PR #123, opened 2023-10-11
fix: a11y issues for attention component

felicia-haggqvist:
Fixes a11y issues raised in [WARP-320 ](https://nmp-jira.atlassian.net/browse/WARP-320) :

- Add role and aria-label to each attention example in Attention.stories.tsx
- Add aria-expanded + aria-controls to popover example to indicate when popover is visible and not. 
- Add aria-describedby, tabIndex to Attention component + add onFocus hook to tooltip example to make tooltip visible on focus. User is now able to tab in order to make the tooltip visible, as well as press `ESC` to close the tooltip:

Skärmavbild 2023-10-11 kl  14 46 01


packages/attention/src/component.tsx
```diff
@@ -137,10 +143,10 @@ const Arrow = forwardRef<HTMLDivElement, ArrowProps>((props, ref) => {
       className={arrowClasses}
       style={{
         // TW doesn't let us specify exactly one corner, only whole sides
-        borderTopLeftRadius: '4px',
+        borderTopLeftRadius: "4px",
         zIndex: 1,
         // border alignment is off by a fraction of a pixel, this fixes it
-        [`margin${arrowDirectionClassname(arrowDirection)}`]: '-0.5px',
+        [`margin${arrowDirectionClassname(arrowDirection)}`]: "-0.5px",
```
imprashast, line 149: Did we change the prettier config or something?

felicia-haggqvist: Could be from my prettier-extension! Do we have a prettier config?

imprashast (approved):

PR #127, opened 2023-10-16
fix: a11y adjust attention

felicia-haggqvist:
Move role + aria-label to the Attention-component
Remove aria-label for Attention-arrow-component


packages/attention/src/component.tsx
```diff
@@ -91,8 +114,8 @@ export function Attention(props: AttentionProps) {
   return (
     <div
       aria-describedby={ariaDescribedby}
-      role={ariaRole}
-      aria-label={ariaLabel}
+      role={props.tooltip ? 'tooltip' : 'img'}
```
BalbinaK, line 117: Just thinking. Would it still be useful to be able to override the `role` and `aria-label` on the whole component? For example, if users want to set `role="tooltip"` on just one child of the Attention, they'd want to reset the default `role` and `aria-label` on the wrapper. We at least displayed an example like this in the Fabric Elements docs: https://elements.fabric-ds.io/attention.html Wdyt?

packages/attention/src/component.tsx
```diff
@@ -91,8 +114,8 @@ export function Attention(props: AttentionProps) {
...
+      aria-label={`${activeAttentionProp()} speech bubble pointing ${pointingAt()}`}
```
BalbinaK, line 118: Just occurred to me that this needs to be translated 🙈

packages/attention/src/component.tsx
```diff
@@ -70,15 +67,41 @@ export function Attention(props: AttentionProps) {
...
+    } else if (actualDirection === "top") {
+      return "down";
+    } else if (actualDirection === "right") {
+      return "left";
+    } else if (actualDirection === "left") {
+      return "right";
+    } else {
+      return "";
+    }
+  }
```
BalbinaK, line 94: `pointingAt` can in fact return `opposites[actualDirection]` and it will result in the same thing.

packages/attention/src/component.tsx
BalbinaK, line 118: I've also just realised that the whole "pointing at ..." should not be added if `props.noArrow === false`.

packages/attention/src/component.tsx
felicia-haggqvist: Ah nice! :D

packages/attention/src/component.tsx
felicia-haggqvist: wait shouldn't it be the opposite? that if we should not add "pointing at..." of props.noArrow === true? Because if don't have an arrow we should say "pointing at..."? :)

felicia-haggqvist: Oh no you're right! 🙈

BalbinaK: Of course, it was supposed to be `props.noArrow === true` 😄

packages/attention/src/component.tsx
felicia-haggqvist: Hmm good point! I could leave the added props for role and aria-label and do a check that if they are empty we will default to what we have set already in the component and otherwise it will take what the user puts in? :)

BalbinaK: Sounds good! Just an empty string should be accepted in case they want to reset our default `role` and `aria-label` 😊

packages/attention/src/component.tsx
felicia-haggqvist: @BalbinaK I have now updated the PR with handling translation for the aria-label. 😄  It would be nice if we wanted to add the functions that I have commented with a `"TODO"` could maybe be moved to the core-repo, but then we would have to handle translations in the core-repo and not sure if we want to do that? 🤔

package.json
```diff
@@ -40,7 +40,7 @@
     "lint:format": "prettier --check . --ignore-path .gitignore",
     "semantic-release": "semantic-release",
     "messages:extract": "lingui extract",
-    "messages:compile": "lingui compile --strict"
+    "messages:compile": "lingui compile"
```
BalbinaK, line 43: Should we remove the `--strict` option forever, or should it be restored when all translations are in place?

packages/_helpers/locales/en/messages.po
```diff
@@ -16,5 +16,5 @@ msgstr ""
 #. Aria label for magnifying glass used as affix in Combobox and Textfield components
 #. js-lingui-explicit-id
 #: packages/_helpers/affix.tsx:37
-msgid "forms.affix.aria.magnifyingGlass"
-msgstr "Magnifying glass"
+#~ msgid "forms.affix.aria.magnifyingGlass"
+#~ msgstr "Magnifying glass"
```
BalbinaK, line 20: I wonder if this change makes anything work differently 🤔

packages/attention/src/component.tsx
```diff
@@ -1,154 +1,237 @@
...
-  rotation,
-  useRecompute as recompute,
-  arrowLabels,
-} from "@warp-ds/core/attention";
-import { attention as ccAttention } from "@warp-ds/css/component-classes";
-import { ArrowProps, AttentionProps } from "./props";
+	opposites,
+	rotation,
+	useRecompute as recompute,
+} from '@warp-ds/core/attention'
```
BalbinaK, line 7: Did we have some setting where the indent was supposed to be so big?

packages/attention/src/component.tsx
```diff
@@ -1,154 +1,237 @@
...
+			message: 'pointing to the',
+			comment:
+				'Default screenreader message for the pointingAt function in the attention component',
+		})
+
+		const pointingAt = !props.noArrow
+			? `${pointingAtTranslation} ${translatedDirection}`
+			: ''
+
+		return `${activeAttentionType} ${pointingAt}`
```
BalbinaK, line 157: Maybe this could be part of the `translatedDirection` constant (it could then be called e.g. `pointingAtDirection` or sth)? Then instead of having:
```
case 'top':
  return .... i18n._({ ... message: 'top', ...}}
```
we'd do:
```
case 'top':
  return .... i18n._({ ... message: ' pointing up', ...}}
```
This would allow us to have this in line 157:
```
return `${activeAttentionType}${props.noArrow ? '' : pointingAtDirection}`
```

Wdyt?

BalbinaK, line 157: Did we skip "speech bubble" as part of the label?

package.json
felicia-haggqvist: @imprashast If I understood correctly from our chat yesterday, we do not need --strict option anymore? :)

packages/_helpers/locales/en/messages.po
felicia-haggqvist: I have no idea, this just updated automatically when I ran `pnpm messages:extract` 🙈

packages/attention/src/component.tsx
felicia-haggqvist: No we need to fix a prettier-config, right now it's just the way my prettier formats the document 🙈

packages/attention/src/component.tsx
felicia-haggqvist: Sure! Sounds good, I'll change it! 😄 👍

felicia-haggqvist: I was thinking if `speech bubble` is maybe not neccessary? Do people understand if we just have `callout` or `tooltip` or `popover`? Otherwise I can put `speech bubble` back and translate it as well. :)

packages/_helpers/locales/en/messages.po
BalbinaK: We can see later if that breaks anything, if now you’re unable to test it due to missing translations 🙂

packages/attention/src/component.tsx
BalbinaK: Could you maybe change that to use 2 spaces instead of tabs? I think that was a configuration we had elsewhere at some point 😅

felicia-haggqvist: Sure! :)

packages/attention/src/component.tsx
BalbinaK: In my opinion, tooltip might be self-explanatory, but popover and callout sound a bit too generic. Is it a popover menu, popover modal etc.? Based on the a11y report, it appears that if a component looks like a speech bubble (💬) it should be described as one.

felicia-haggqvist: That's true! Ok I will add `speech bubble` then! :)

packages/attention/stories/Attention.stories.tsx
```diff
@@ -1,48 +1,62 @@
...
-        ariaLabel="Green speech bubble pointing left"
-        ariaRole="img"
+        role=''
+        aria-label=''
+        placement='right'
         isShowing={true}
       >
-        <p style={{ width: 200 }}>
-          I'm a callout because that box over there is new or something
+        <p id="callout-reset-bubbletext" style={{ width: 200 }}>
```
BalbinaK, line 36: I _think_ that in order for this to work, you'd need to set `role='tooltip'` or `role='img'` on this element. Otherwise the screen reader won't know that it should read this text. I'd also add `pointing left` at the end of the example in line 37, just to display that we should not forget to explain what the component looks like.

felicia-haggqvist: Ah you're probably right! I guess I should add `role='tooltip`' or `role='img'` for the `p`-element in the reset-example in the vue-PR as well then? :)

BalbinaK (approved):

PR #128, opened 2023-10-17
fix: ay11 fixes for the Modal component and affected stories

AnnaRybkina:
Background: https://nmp-jira.atlassian.net/jira/software/projects/WARP/boards/15?label=web&selectedIssue=WARP-321

Fixes in the component: 
- exchange `p` with `h1`
- if close button - set focus on it instead of the first interactive element. This change needs to be reflected in the docs. See 👇 
image

Fixes in stories:
- focus on the "Open modal" button
- reset checked on modal close
- initial focus visible

### DIFF
--- a/packages/modal/src/component.tsx
+++ b/packages/modal/src/component.tsx
@@ -22,6 +22,7 @@ export const Modal = ({
 }: ModalProps) => {
   const contentRef = useRef<HTMLDivElement>(null);
   const id = useId(props.id);
+  const closeButtonRef = useRef<HTMLButtonElement>(null);
 
   activateI18n(enMessages, nbMessages, fiMessages);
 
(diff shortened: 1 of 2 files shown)


packages/modal/src/component.tsx
```diff
@@ -32,9 +33,12 @@ export const Modal = ({
...
   useEffect(() => {
-    if (!props.initialFocusRef) return;
-    props.initialFocusRef.current?.focus();
-  }, [props.open, props.initialFocusRef]);
+    if (!props.initialFocusRef) {
+      props.right && closeButtonRef.current?.focus();
+    } else {
+      props.initialFocusRef.current?.focus();
+    }
+  }, [props.open, props.initialFocusRef, closeButtonRef]);
```
BalbinaK, line 41: Do we need `closeButtonRef` in the dependency array? If so, should we add `props.right`, too?

BalbinaK (approved):

packages/modal/src/component.tsx
AnnaRybkina: ah, no you are right, that can be removed. And the `props.right` should be added instead. Don't think it is really necessary though but it can change and re-render the component

AnnaRybkina: https://github.com/warp-ds/react/pull/128/commits/54f4c5f48a1da68a12d584bbc840a6dc799aa101

packages/modal/src/component.tsx
```diff
@@ -109,14 +113,15 @@ export const Modal = ({
               )}
             >
               {typeof props.title === "string" ? (
-                <p className={ccModal.titleText}>{props.title}</p>
+                <h1 className={ccModal.titleText}>{props.title}</h1>
```
BalbinaK, line 116: 🎉

felicia-haggqvist (approved):

PR #136, opened 2023-10-20
chore: try to build preview storybook

AnnaRybkina:
Triggar deploy


.github/workflows/preview.yml
```diff
@@ -0,0 +1,36 @@
...
+        uses: pnpm/action-setup@v2
+        with:
+          version: 8
+          run_install: true
+      - name: Build storybook
+        run: pnpm run build-storybook
+      - name: Deploy preview
+        uses: rossjrw/pr-preview-action@v1
+        with:
+          source-dir: storybook-static
```
AnnaRybkina, line 33:
```suggestion
          source-dir: ./storybook-static/
```

.github/workflows/preview.yml
```diff
@@ -0,0 +1,36 @@
+name: Storybook preview
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - closed
```
AnnaRybkina, line 9:
```suggestion
      - closed
      - edited
```

.github/workflows/preview.yml
```diff
@@ -0,0 +1,37 @@
...
+      - edited
```
BalbinaK, line 10: Why do we need to redeploy when PR info is edited? Here the author of the GH action mentions that we should run the action only when files change: https://github.com/rossjrw/pr-preview-action#run-only-when-files-are-changed

AnnaRybkina: This is for trigger the deployment. For my testing. Should be removed before merging. I added it ONLY for triggering the deployment. I should have had `revert` this but it was late and I didn't think about it. You read correctly

AnnaRybkina: So, if we do changes in this PR the workflow won't trigger without this thing.

BalbinaK (approved):
I think this is a great feature! Amazing that you looked into how to facilitate our review process! 
Maybe one more question here. It could be that some PRs will not introduce any changes to our components, or will not require previews for proper review. Would it maybe make sense to have such previews deployed only on demand? Not sure if it's easy to implement, so it's more of an open question.

.github/workflows/preview.yml
AnnaRybkina, line 10: This is only added for testing aka trigger deployment if we do changes in this workflow in this PR. IMPORTANT to remove before merging
```suggestion
```

AnnaRybkina: Added comment here https://github.com/warp-ds/react/pull/136/files#r1368234462

BalbinaK (approved):

PR #138, opened 2023-10-23
fix: a11y issues for expandable

felicia-haggqvist:
Fix a11y issues for expandable component in [WARP-326](https://nmp-jira.atlassian.net/browse/WARP-326) :

- Switch between ChevronDownIcon and ChevronUpIcon depending on if expanded is set to true or not. This will set the correct svg-title for the icons. 
- Move Icon in the "Box With Custom Icon" story to after the text.
- Add a comma with sr-only to add a pause between the text and the svg-title for screen readers.
- Change text "I am expandable" to a `<h2>` instead of a `<h1>`

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
     "@chbphone55/classnames": "^2.0.0",
     "@lingui/core": "^4.3.0",
     "@warp-ds/core": "^1.0.0",
-    "@warp-ds/css": "^1.1.3-next.1",
+    "@warp-ds/css": "^1.2.0",
     "@warp-ds/uno": "^1.1.0",
     "@warp-ds/icons": "1.1.1",
     "react-focus-lock": "^2.5.2",
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
     "@chbphone55/classnames": "^2.0.0",
     "@lingui/core": "^4.3.0",
     "@warp-ds/core": "^1.0.0",
-    "@warp-ds/css": "^1.1.2",
+    "@warp-ds/css": "^1.1.3-next.1",
     "@warp-ds/uno": "^1.1.0",
     "@warp-ds/icons": "1.1.0",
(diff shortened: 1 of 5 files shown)


magnuh:
Looks like you are using tabs instead of spaces for indentation? 😅

packages/expandable/src/props.tsx
```diff
@@ -32,6 +32,12 @@ export type ExpandableProps = {
    */
   expanded?: boolean;
 
+  /**
+   * The state of the chevron-up icon, if set to true, chevron up should be displayed, if set to false then chevron-down icon should be displayed.
+   * @default false
+   */
+  showChevronUpIcon?: boolean;
```
BalbinaK, line 39: Why can't we use `expanded` for that purpose? If it's `true`, chevron down should be displayed and the opposite.

package.json
```diff
@@ -91,7 +91,7 @@
     "@chbphone55/classnames": "^2.0.0",
     "@lingui/core": "^4.3.0",
     "@warp-ds/core": "^1.0.0",
-    "@warp-ds/css": "^1.1.2",
+    "@warp-ds/css": "^1.1.3-next.1",
```
BalbinaK, line 94: When we get this out we should be able to bump the version to 1.2.0: https://github.com/warp-ds/css/pull/75

BalbinaK (approved):

packages/expandable/stories/Expandable.stories.tsx
```diff
@@ -57,31 +58,37 @@ export const NoChevron = () => {
...
+      <h2>I am expandable</h2>
     </Expandable>
-  );
-};
+  )
+}
 
 export const Heading = () => {
   return (
     <Expandable title="I'm also a heading" headingLevel={1}>
```
BalbinaK, line 90: this needs to be changed to 2

BalbinaK (approved):

PR #140, opened 2023-10-23
fix: add a11y attributes for to define steps in more accessible way

AnnaRybkina:
Background: https://nmp-jira.atlassian.net/jira/software/projects/WARP/boards/15?label=web&selectedIssue=WARP-329

- [x] Added role=img and aria-label on the divs that renders the circles (instead of alt text)
- [x] added list markup to communicate the total number of steps to the screen reader
- [x] Added aria-current=step to indicate current step

image

### DIFF
--- a/packages/steps/src/locales/en/messages.mjs
+++ b/packages/steps/src/locales/en/messages.mjs
@@ -1,0 +1,1 @@
+/*eslint-disable*/export const messages=JSON.parse("{\"steps.aria.emptyCircle\":\"Empty circle\",\"steps.aria.active\":\"Step indicator active circle\",\"steps.aria.completed\":\"Step indicator completed circle\"}");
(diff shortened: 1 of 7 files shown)


packages/steps/src/locales/en/messages.mjs
```diff
@@ -0,0 +1 @@
+/*eslint-disable*/export const messages=JSON.parse("{\"steps.aria.emptyCircle\":\"Empty circle\",\"steps.aria.active\":\"Step indicator active circle\",\"steps.aria.completed\":\"Step indicator completed circle\"}");
```
imprashast, line 1: We should not checkin this file

imprashast (approved):

BalbinaK (approved):

PR #142, opened 2023-10-24
fix: crowdin build step

felicia-haggqvist:
This should fix the failed synchronize-with-crowdin step that is currently on next. 

After speaking to Christoffer I found out that there's currently a bug in Crowdin, but that they are working on to fix. In the meanwhile I have the updated ReadMe-file with temporary solution to fix this issue.

### DIFF
--- a/packages/_helpers/locales/fi/messages.po
+++ b/packages/_helpers/locales/fi/messages.po
@@ -17,9 +17,3 @@ msgstr ""
 "X-Crowdin-Language: fi\n"
 "X-Crowdin-File: /react/packages/_helpers/messages.po\n"
 "X-Crowdin-File-ID: 651\n"
-
-#. Aria label for magnifying glass used as affix in Combobox and Textfield components
-#. js-lingui-explicit-id
-#: packages/_helpers/affix.tsx:37
-#~ msgid "forms.affix.aria.magnifyingGlass"
-#~ msgstr "Suurennuslasi"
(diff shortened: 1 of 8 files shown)


imprashast (approved):

packages/_helpers/locales/fi/messages.po
```diff
@@ -17,9 +17,3 @@ msgstr ""
 "X-Crowdin-Language: fi\n"
 "X-Crowdin-File: /react/packages/_helpers/messages.po\n"
 "X-Crowdin-File-ID: 651\n"
-
-#. Aria label for magnifying glass used as affix in Combobox and Textfield components
-#. js-lingui-explicit-id
-#: packages/_helpers/affix.tsx:37
-#~ msgid "forms.affix.aria.magnifyingGlass"
-#~ msgstr "Suurennuslasi"
```
BalbinaK, line 25: I don't know if that could be the reason, but it looks like the clean script removes whatever is using `#~`. I wonder if making the two last lines the same as in other .po files, so without the `#~`, would solve the issue. Otherwise I don't understand how the translations can work if we remove these lines completely 😅

felicia-haggqvist: but the .po files are automatically generated by running `lingui extract` so how should we manually change how they are being generated? :)

BalbinaK: I don't understand how it works that some files have `#~` before  `msgid` and some don't while running `lingui extract` (it felt like the missing `--strict` option in that command had something to do with it). I merely noticed that it seemed those `#~` characters might be the reason that the crowdin sync job failed. But we should be able to check it by merging this PR I guess 🙂

felicia-haggqvist: Yes I'm hoping that this solves the crowdin sync job failed, since this is the solution that Christoffer suggested to implement to fix the build. So hopefully it should work! 🙂

PR #144, opened 2023-10-25
fix: activate translations in steps

AnnaRybkina:
I guess I forgot to do these changes in https://github.com/warp-ds/react/pull/140


BalbinaK (commented):
Should we add steps to `./crowdin.yml`?

AnnaRybkina:
> Should we add steps to `./crowdin.yml`?

ah true, good catch

BalbinaK (approved):

PR #145, opened 2023-10-25
feat(attention): add highlight variant

BalbinaK:
Fixes [WARP-275](https://nmp-jira.atlassian.net/browse/WARP-275)

Screenshot 2023-10-25 at 16 44 31


packages/attention/stories/Attention.stories.tsx
```diff
@@ -41,6 +41,44 @@ export function CalloutResettingRoleAndAriaLabel() {
...
+          <Box info>
+            <h1>I am a box full of info</h1>
+          </Box>
+        </div>
+        <Attention 
+          highlight
+          placement='top'
+          isShowing={show}
+          targetEl={targetEl}
+          id='highlight-attention-example'
```
felicia-haggqvist, line 71: I think the id needs to be added to the component.tsx file in the wrapperClasses-div. Otherwise I cannot find this id on the Attention-component when inspecting the DOM. :)

BalbinaK: Ah, good catch! Fixed in [78ed95a](https://github.com/warp-ds/react/pull/145/commits/78ed95a9950706aabedb14e2a433c436e30c5d9f)

felicia-haggqvist (approved):

PR #150, opened 2023-11-02
fix: a11y issues for pill

felicia-haggqvist:
Solves this Jira-ticket: [WARP-332](https://nmp-jira.atlassian.net/browse/WARP-332)
- Remove `sr-only` text + sr-only props from pill
- Remove locales-folder for pill
- add `aria-label` prop to pill + add `aria-label` to pillsWithIcon story


packages/pill/stories/Button.stories.tsx
```diff
@@ -1,54 +1,81 @@
...
+}
 
 export const PillsWithIcon = () => {
   return (
-    <div className="flex space-x-8">
+    <div className='flex space-x-8'>
       <Pill
+        aria-label='a pill with a search icon'
         icon={<IconSearch16 />}
         onClick={() => alert('onClick event')}
```
BalbinaK, line 69: In the [ticket](https://nmp-jira.atlassian.net/browse/WARP-332) it was mentioned that we should use role="img" and aria-label instead of `title`, which I _think_ was regarding the use of Pill with icon in storybook. I don't think we have the possibility to hide `title` element from the svg at this point so we can just have role and aria-label on the Pill.

packages/pill/src/component.tsx
```diff
@@ -1,67 +1,41 @@
...
+import { PillProps } from '.'
+import { pill as ccPill } from '@warp-ds/css/component-classes'
 
 export function Pill(props: PillProps) {
-  activateI18n(enMessages, nbMessages, fiMessages);
   return (
-    <div className={ccPill.pill}>
+    <div
+      role='img'
+      aria-label={props['aria-label'] === '' ? undefined : props['aria-label']}
```
BalbinaK, line 10: Now that I think about it, `role="img"` should always be combined with an attribute that describes such "image". Here we'd use `aria-label` but we know it can be `undefined`. I think this component should only accept `role` as props, in case you want to describe contents of the Pill with custom `aria-label`. 

Based on the [Jira ticket](https://nmp-jira.atlassian.net/browse/WARP-332) I see the accessibility team suggested that we use `role=img` and `aria-label` instead of `title`, which I think referred only to the story `PillsWithIcon`, and not to all Pill instances.

felicia-haggqvist: I think we still need to have `aria-label` as a prop on pill otherwise it will not understand what `aria-label` is when using it on the `<Pill>` tag.

BalbinaK: Sounds good!

packages/pill/src/component.tsx
```diff
@@ -6,7 +6,7 @@ import { pill as ccPill } from '@warp-ds/css/component-classes'
 export function Pill(props: PillProps) {
   return (
     <div
-      role='img'
+      role={props.role === '' ? undefined : props.role}
```
BalbinaK, line 9: Isn't it enough to have `role={props.role}` here? Same with `aria-label` in the next line. The reason why we had such check in other components that accepted aria-label and role properties was because we'd default to some value that should be possible to reset/override. Here we're not defaulting to anything, so I think we can make it simple :)

felicia-haggqvist: ah ok! That makes sense! :)

BalbinaK (approved):

PR #158, opened 2023-11-10
fix(combobox): open dropdown also on space key press

BalbinaK:
Fixes [WARP-333](https://nmp-jira.atlassian.net/browse/WARP-333)

This makes Combobox dropdown open on 'Space' key press, like in the case of the Select component. This was the suggestion from the a11y team. The proper solution should be to open the dropdown on Ctrl+Option+Space combination press. However VoiceOver doesn't detect this key combination for some reason, so having a the dropdown to open on `Space` key press felt like a good enough solution for now.


BalbinaK:
Feedback from a11y team: "The info from VO about pressing ctrl+opt+space is excessive because the list will open anyway when people type. Since it's not a conventional select box with a fixed number of suggestions, it makes no sense to open it before the user has typed something."

PR #169, opened 2023-11-22
fix: Import icons in the correct way

imprashast:
Switch to importing individual icons since that was polluting the bundle. But this is just step 1. It might be that we need to publish individual exports like we do in elements repo.

Currently using icons: `1.3.0-next.2`


felicia-haggqvist:
Shouldn't this wait until there is a stable new release of icons and use that instead of the latest next-release? I thought we were supposed to avoid adding the "-next" releases to our component-repos, but maybe I've minsunderstood that. :)

felicia-haggqvist (approved):

BalbinaK (commented):
Should the `meta.json` file be added to .gitignore? What is it used for?

imprashast:
Ah, the `meta.json` is used for bundle analysis. I'll remove that file and show that in my Show & Tell for Tree-shaking :)

PR #173, opened 2023-11-23
fix(textfield): allow styling input's left padding if prefix is wider than 40px

BalbinaK:
Fixes [WARP-374](https://nmp-jira.atlassian.net/browse/WARP-374)

Prefix is an absolutely positioned component and the gap between that component and the `input`'s content is handled using a left padding on the `input`. It’s currently hard-coded to 40px, but with this quick fix we can let users change it by setting a CSS variable in `TextField`’s class (`<TextField class=”[--w-prefix-wdith: 100px]”>`). It’s not dynamic and users will need to calculate how much padding their input needs, but it’s the only non-breaking quick fix we managed to come up with. We should really consider rewriting these form components completely to properly approach styling of different text fields, affixes etc.

Screenshot with long prefix
Screenshot with 2-letter prefix
Screenshot with no prefix

### DIFF
--- a/packages/textfield/src/component.tsx
+++ b/packages/textfield/src/component.tsx
@@ -46,6 +46,14 @@ export const TextField = forwardRef<HTMLInputElement, TextFieldProps>(
           className={className}
           style={style}
         >
+          {/* we style input with prefix here because we cannot use arbitrary values with commas in UnoCSS like pl-[var(--w-prefix-width, 40px)] */}
+          <style>
+            {`
+              div+#${id}, button+#${id} {
+                padding-left:var(--w-prefix-width, 40px);
+              }
+            `}
+          </style>
           {label && (
             <label htmlFor={id} className={classNames({
               [ccLabel.label]: true,
--- a/packages/textfield/stories/Textfield.stories.tsx
+++ b/packages/textfield/stories/Textfield.stories.tsx
@@ -50,6 +50,12 @@ export const labelPrefix = () => (
   </TextField>
 );
 
+export const longLabelPrefix = () => (
+  <TextField className="[--w-prefix-width:90px]" value="With some value">
+    <Affix prefix label="Long prefix" />
+  </TextField>
+);
+
 export const clearSuffix = () => (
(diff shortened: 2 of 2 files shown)


packages/textfield/stories/Textfield.stories.tsx
```diff
@@ -50,6 +50,12 @@ export const labelPrefix = () => (
   </TextField>
 );
 
+export const longLabelPrefix = () => (
+  <TextField className="[--w-prefix-width:90px]" value="With some value">
```
imprashast, line 54: Does this mean teams using this would have to define width if the text is longer themself and it won't happen dynamically based on content length?

packages/textfield/src/component.tsx
```diff
@@ -46,6 +46,14 @@ export const TextField = forwardRef<HTMLInputElement, TextFieldProps>(
           className={className}
           style={style}
         >
+          {/* we style input with prefix here because we cannot use arbitrary values with commas in UnoCSS like pl-[var(--w-prefix-width, 40px)] */}
```
BalbinaK, line 49: Tailwind had a similar issue: https://github.com/tailwindlabs/tailwindcss-jit/pull/91

packages/textfield/stories/Textfield.stories.tsx
BalbinaK: That's correct. They will only need to do so if their label is wider than the default 40px though (sounds like 1 team so far). Calculating width of an element from the component's perspective might be another temporary solution, but we didn't really have a proper testing environment to quickly investigate if such solution would work in most cases (SSR, Shadow DOM etc.). We've also already spent quite some time trying to find any quick fix for this small bug that wouldn't be a breaking change, so we felt we shouldn't spend more time on yet another temporary fix. That is also because we strongly believe this component needs a different approach, which would among other things make use of flexbox. That would however require a rewrite that we cannot afford right at this point. Let me know what you think! 😊

imprashast (approved):

PR #184, opened 2023-12-15
fix: Improve exception handling in Tabs

magnuh:
Removing try/catch blocks and improving exception handling in Tabs in an attempt to relieve symptoms described in this thread: https://sch-chat.slack.com/archives/C04P0GYTHPV/p1702545068719299

image

### DIFF
--- a/packages/tabs/src/component-tabs.tsx
+++ b/packages/tabs/src/component-tabs.tsx
@@ -1,4 +1,4 @@
-import {
+import React, {
   useState,
   useRef,
   useEffect,
--- a/packages/tabs/src/component-tabs.tsx
+++ b/packages/tabs/src/component-tabs.tsx
@@ -1,9 +1,10 @@
-import React, {
+import {
   useState,
   useRef,
   useEffect,
   cloneElement,
   Children,
+  RefObject,
 } from 'react';
 import { classNames } from '@chbphone55/classnames';
 import { gridLayout, tabs as ccTabs } from '@warp-ds/css/component-classes';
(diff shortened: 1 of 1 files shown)


BalbinaK (commented):
You mentioned in the linked thread that this will probably not fix the underlying problem the user had but will remove the symptoms (logged warning). The way I see it, there is a problem with types in the screenshot, and not the try-catch you removed. Why did you decide to remove it? 
Also, even if we go for such change, I'm afraid we'll need to make a similar adjustment in Vue to align between the implementations of Tabs :)

magnuh:
> You mentioned in the linked thread that this will probably not fix the underlying problem the user had but will remove the symptoms (logged warning). The way I see it, there is a problem with types in the screenshot, and not the try-catch you removed. Why did you decide to remove it? Also, even if we go for such change, I'm afraid we'll need to make a similar adjustment in Vue to align between the implementations of Tabs :)

There is no problem with "types", there is a problem with not having proper exception handling... which then has been "solved" by adding try/catch blocks around error-prone code. I've handled what potentially would throw exceptions... But the underlying problem is probably something related to React hooks and rendering/re-rendering of the component. And that would be a bigger task. The component "works", but it threw exceptions which only resulted in console.warns in prod, which is not a nice thing, nor the way that one should handle exceptions.

magnuh:
My guess is that we won't have the same "underlying problem" in Vue... But yes, we should of course not have that try/catch exception "handling" there either.

BalbinaK:
> > You mentioned in the linked thread that this will probably not fix the underlying problem the user had but will remove the symptoms (logged warning). The way I see it, there is a problem with types in the screenshot, and not the try-catch you removed. Why did you decide to remove it? Also, even if we go for such change, I'm afraid we'll need to make a similar adjustment in Vue to align between the implementations of Tabs :)
> 
> There is no problem with "types", there is a problem with not having proper exception handling... which then has been "solved" by adding try/catch blocks around error-prone code. I've handled what potentially would throw exceptions... But the underlying problem is probably something related to React hooks and rendering/re-rendering of the component. And that would be a bigger task. The component "works", but it threw exceptions which only resulted in console.warns in prod, which is not a nice thing, nor the way that one should handle exceptions.

Ah, I see now! I thought there was more to the try-catch statement and that confused me a bit.

BalbinaK (approved):

packages/tabs/src/component-tabs.tsx
```diff
@@ -1,9 +1,10 @@
-import React, {
```
BalbinaK, line 1: When this is removed, TS complains about all returned jsx elements: 
```
'React' refers to a UMD global, but the current file is a module. Consider adding an import instead. ts(2686) 
```
Also, the `pnpm build` fails.

BalbinaK (changes requested):

BalbinaK (approved):

PR #185, opened 2023-12-19
fix(card): set default background on unselected non-flat card

BalbinaK:
Fixes [WARP-430](https://nmp-jira.atlassian.net/browse/WARP-430)

Background color of a on-flat Card will no longer be inherited from the parent container.

Before:
Screenshot 2023-12-19 at 13 38 04

After:
Screenshot 2023-12-19 at 15 07 38

### DIFF
--- a/packages/card/src/component.tsx
+++ b/packages/card/src/component.tsx
@@ -20,7 +20,7 @@ export function Card(props: CardProps) {
       className: classNames(props.className, {
         [ccCard.card]: true,
         [ccCard.cardShadow]: !props.flat,
-        [ccCard.cardSelected]: props.selected,
+        [props.selected ? ccCard.cardSelected : ccCard.cardUnselected]: !props.flat,
         [ccCard.cardFlat]: props.flat,
         [props.selected ? ccCard.cardFlatSelected : ccCard.cardFlatUnselected]:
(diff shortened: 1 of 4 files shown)


packages/card/src/component.tsx
```diff
@@ -20,7 +20,7 @@ export function Card(props: CardProps) {
       className: classNames(props.className, {
         [ccCard.card]: true,
         [ccCard.cardShadow]: !props.flat,
-        [ccCard.cardSelected]: props.selected,
+        [props.selected ? ccCard.cardSelected : ccCard.cardUnselected]: !props.flat,
```
magnuh, line 23: The flat variant would never get the selected state? 🤔

magnuh: But maybe it shouldn't, but did before? 😅

BalbinaK: It wouldn't? Then why 25 handles selected/unselected state for flat card? :D

magnuh: Before, a selected flat card would have got both `ccCard.selected` (row 23) and `ccCardFlatSelected` (row 25)? Right? 🤷😄

magnuh (approved):

PR #187, opened 2023-12-28
fix: add header classes modal

felicia-haggqvist:
**Note:** Once [this PR](https://github.com/warp-ds/css/pull/116) has been merged and we have published a new version of @warp-ds/css, we need to update the version of @warp-ds/css in the package.json before this PR can be merged.

These changes are part of fixes for this JIRA ticket: [WARP-429](https://nmp-jira.atlassian.net/browse/WARP-429)

- Added `headerClasses` prop to Modal + added it to modal-example, changing the height of the modal title to `h-full sm:h-full`.
- Adjusted styling for the modal title after changes made in [this PR](https://github.com/warp-ds/css/pull/116)

**To test:** Link this branch with branch from the css-repo: `fix/refactor-modal-style`

### DIFF
--- a/packages/modal/src/component.tsx
+++ b/packages/modal/src/component.tsx
@@ -111,8 +111,9 @@ export const Modal = ({
               id={`${id}__title`}
               className={classNames(
                 ccModal.transitionTitle,
+                ccModal.transitionTitleMaxWidth,
                 !!props.left
-                  ? ccModal.transitionTitleCenter
+                  ? [ccModal.transitionTitleCenter, ccModal.transitionTitleColSpan]
                   : ccModal.transitionTitleColSpan
               )}
             >
--- a/packages/modal/src/component.tsx
+++ b/packages/modal/src/component.tsx
@@ -75,7 +75,10 @@ export const Modal = ({
           className={ccModal.modal}
           tabIndex={-1}
         >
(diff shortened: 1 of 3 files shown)


Skadefryd:
I think we might be just moving the problems arround on this? Height of the top should just be following the length of the content in there, but theres also other issues that might be out of scope for this pr? Just to mention some:  spacing is out of balance, that line is weirdly implemented, it could(prob should) only show up  when there is a overflow so the scroll has something to 'scroll into' visually. Scroll should end on the line not on the line +16px. Line should be both over and under (maybe hide top line when you are on top of the scroll, and the bottom one when on the bottom of the scroll?). I dont know why but the line seams to be implemented mobile only instead of on overflow.  Bottom buttons seams to be weirly placed, especially on smaller screens.

felicia-haggqvist:
@Skadefryd All good points and I agree that this PR has grown out of scope! 👍  Maybe for this PR we should just remove the fixed height on the title and then look into how to refactor the behavior of the divider underneath the title + repositioning the left and right buttons for another ticket? Feels like we need to sit down and figure out the best way to solve the modal design, because I agree that now it just feels like we are just moving the problem. 🙈

felicia-haggqvist:
Will close this PR, due to that the ticket has grown out of scope.

PR #189, opened 2024-01-05
feat: Add Pagination component to WARP

imprashast:
This will add Pagination Component to WARP [WARP-135](https://nmp-jira.atlassian.net/browse/WARP-135)

Component on Desktop (From storybook):
Screenshot 2024-01-05 at 16 09 52

Component on Mobile (From storybook):
Screenshot 2024-01-05 at 16 10 06


package.json
```diff
@@ -90,10 +90,10 @@
   "dependencies": {
     "@chbphone55/classnames": "^2.0.0",
     "@lingui/core": "^4.3.0",
-    "@warp-ds/core": "^1.0.0",
-    "@warp-ds/css": "^1.6.1",
-    "@warp-ds/uno": "^1.1.0",
-    "@warp-ds/icons": "1.3.0-next.2",
+    "@warp-ds/core": "^1.0.2",
+    "@warp-ds/css": "1.7.0-next.1",
```
imprashast, line 94: Need to release warp-ds/css before merging

packages/pagination/src/Page.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  return (
+    <Button
+      pill
+      aria-label={ariaLabel}
+      {...props}
+      ref={ref}
+      rel="nofollow"
+      aria-current={isCurrentPage ? 'page' : undefined}
+      className={classNames(className, ccPagination.defaultPage, {
```
BalbinaK, line 56: I see that on hover the background color changes to `s-bg-primary-subtle-hover`, which looks like this:
Screenshot 2024-01-08 at 12 31 09
Shouldn't the colors get darker on hover and active state like in the case of primary buttons? I couldn't see any design examples for pagination component so I thought I'd double-check here.

imprashast: The colors are from what @magnuh has in pagination.yml (https://github.com/warp-ds/css/pull/121/commits/dae034881766d9cf26247ce34dc3ff000a0dd20b)

packages/pagination/src/CurrentPage.tsx
```diff
@@ -0,0 +1,44 @@
...
+  children?: React.ReactNode;
+} & React.HTMLAttributes<HTMLDivElement>;
+
+const CurrentPage = React.forwardRef<HTMLDivElement, CurrentPageProps>(
+  ({ children, className, ...props }, ref) => {
+    const { currentPage } = usePagination();
+
+    if (currentPage <= 1) {
+      return null;
+    }
```
BalbinaK, line 20: I think a component should always render something - I would move this condition to the parent instead, so it's apparent already there what will and what won't be rendered. Wdyt?

packages/pagination/src/FirstPage.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+const FirstPage = React.forwardRef<
+  React.AnchorHTMLAttributes<HTMLAnchorElement>,
+  FirstPageProps
+>(({ className, ...props }, ref) => {
+  const { currentPage } = usePagination();
+
+  if (currentPage <= 2) {
+    return null;
+  }
```
BalbinaK, line 35: Same here.

packages/pagination/src/FirstPage.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { pagination as ccPagination } from '@warp-ds/css/component-classes';
+import { usePagination } from './PaginationContainer.js';
+import IconChevronDoubleLeft16 from '@warp-ds/icons/react/chevron-double-left-16';
+import React from 'react';
+
+type FirstPageProps = {
+  /**
+   * @default First page
+   */
+  'aria-label'?: string;
```
BalbinaK, line 12: I think you should use `aria-label` or the translated default in line 37

packages/pagination/src/FirstPage.tsx
```diff
@@ -0,0 +1,58 @@
...
+  });
+
+  return (
+    <Button
+      pill
+      ref={ref}
+      {...props}
+      aria-label={ariaLabel}
+      className={className}
+      rel="start"
```
BalbinaK, line 51: What will `rel="start"` do here?

packages/pagination/src/FirstPage.tsx
imprashast: What do you mean? For someone using the FirstPage.tsx, they can either suppy their own aria-label or use the one we have in this library

packages/pagination/src/FirstPage.tsx
imprashast: This used to be for SEO, to specify which is the starting point but I'm not sure if it's needed

packages/pagination/src/FirstPage.tsx
BalbinaK: I meant that you should maybe do what you did in the PaginationContainer here: https://github.com/warp-ds/react/pull/189/files#diff-b4a4e56035a870c1373c2bf17afc691e9fdd9321e42da35f7a913d36ef3cbf71R67-R74
This way if aria-label prop exists, it would be used instead of the default one.

packages/pagination/src/NextPage.tsx
```diff
@@ -0,0 +1,86 @@
...
+          <span className={ccPagination.firstPageLabel}>{label}</span>
+        </Button>
+      )}
+      <Button
+        pill
+        aria-label={ariaLabel}
+        {...props}
+        ref={ref}
+        rel="next nofollow"
+        className={classNames(ccPagination.nextPage, className)}
```
BalbinaK, line 78: Maybe that's intended, but I saw that on mobile when I press Next/PrevPage button, the hover background color remains visible until I click elsewhere.
pagination-mobile
Screenshot 2024-01-08 at 13 23 13

packages/pagination/src/Page.tsx
magnuh: The colors are from the tokens sheet: https://docs.google.com/spreadsheets/d/1Q-Tr_dwJVfxgh3527IFjXQKqtnPPM42QeIr-M6q-zu8/edit#gid=1588196525&range=293:293
...so that is the correct behaviour. 🙂 
...but shouldn't the text for non active pages be link color (blue)? 🤔

BalbinaK: I now found a design [in Figma](https://www.figma.com/file/oHBCzDdJxHQ6fmFLYWUltf/Adi-sandbox%3A-Warp-variables-v2.0?type=design&node-id=818-35572&mode=design&t=LRe5aAQmwbCo9xYv-0), which also shows the primary-subtle-hover color.

felicia-haggqvist: It also looks like in the design from Figma that the text should have a font-weight 300, while here it seems to have a font-weight of 700. 🙂

imprashast: Hmm, we don't support 300 font weight. Only 400 & 700

packages/pagination/src/NextPage.tsx
imprashast: It doesn't matter since the page would have changed/reloaded. You can see this on the result page on tori/finn

packages/pagination/src/Page.tsx
felicia-haggqvist: Oh! 😄

BalbinaK (approved):
LGTM

felicia-haggqvist (approved):

PR #190, opened 2024-01-10
feat: add WIP testing setup

imprashast:
Opening this PR to start a discussion around testing our components


trygve-lie:
Would operating with an test suite which has generic tests and we apply this to the different frameworks be sensible here.

I mean; ideally a component should exist and behave similar in all frameworks (generate the same markup etc), can we have one test and have it run on different frameworks. That will also unify behavior between frameworks.

trygve-lie:
Any consideration of using [Playwright](https://playwright.dev/) over doing a node implementation of the DOM?

PR #193, opened 2024-01-18
chore: Initial unit testing setup

imprashast:
Based on our discussions, starting the initial setup of unit testing in our react app. Similar setup will be added to the vue and elements apps.


imprashast:
Marking this as a chore since we don't want this to trigger a release.

tests/components/PaginationTest.tsx
```diff
@@ -0,0 +1,38 @@
...
+            onChange={onChangeFunction} />);
+    });
+
+    it('renders', () => {
+        expect(screen.getByText('Pages')).toBeInTheDocument();
+    });
+
+    it('renders an H1', () => {
+        expect(screen.getByRole('heading', { level: 1 })).toBeInTheDocument();
+        expect(screen.getByRole('heading', { level: 1 })).toHaveTextContent('Pages');
```
magnuh, line 24: Would it work to chain matchers? 🤔 
```suggestion
        expect(screen.getByRole('heading', { level: 1 })).toBeInTheDocument().and.toHaveTextContent('Pages');
```
...and would it be better/faster? 🤷😄

imprashast: No it doesn't. The counter argument again chaining assertions is that when something fails, it's easier to see which line it fails on. Would be harder to debug when statements are chained, is it the first one or the second one?

magnuh: Agreed. 🙂👍  ...but then, how about this? 😄 
```suggestion
        const heading = screen.getByRole('heading', { level: 1 });
        expect(heading).toBeInTheDocument();
        expect(heading).toHaveTextContent('Pages');
```

imprashast: Sure 👍

magnuh (approved):
🥳

felicia-haggqvist (approved):

package.json
```diff
@@ -58,12 +60,15 @@
...
     "@storybook/react-vite": "^7.6.9",
+    "@testing-library/jest-dom": "^6.2.0",
+    "@testing-library/react": "^14.1.2",
     "@types/node": "^20.11.5",
     "@types/react": "^17.0.75",
     "@types/react-dom": "^17.0.25",
     "@typescript-eslint/eslint-plugin": "^6.19.0",
     "@typescript-eslint/parser": "^6.19.0",
     "@vitejs/plugin-react": "^4.2.1",
+    "@vitest/coverage-v8": "^1.2.1",
```
BalbinaK, line 71: Is this used anywhere?

PR #195, opened 2024-01-23
fix: SSR compatibility for textfield

imprashast:
Before this change, the value of `id` from the useId hook would initialize with null on server and get a value on client which broke SSR hydration. See screenshot of before:
Screenshot 2024-01-23 at 11 23 07

To fix this, we can use the `useId` hook from react which generates this value on server as well, hence no mismatch between server and client. See screenshot below:
Screenshot 2024-01-23 at 11 22 40

Notice how the value was printed twice before and only once after. This was because the our homemade useId hook ran only clientside with the useEffect hook.


packages/textfield/src/component.tsx
```diff
@@ -28,8 +27,10 @@ export const TextField = forwardRef<HTMLInputElement, TextFieldProps>(
     } = props;
 
     activateI18n(enMessages, nbMessages, fiMessages);
-
-    const id = useId(providedId);
+    
+    // useId returns a string that includes colons (:), e.g., :r0:, :r1:, etc.
+    // This string is NOT supported in CSS selectors. Hence the replace.
+    const id = providedId ?? useId().replace(/:/g, '');
```
BalbinaK, line 33: Not sure why the original `useId` util had to rely on `useLayoutEffect`, but would it make sense to move the React `useId()` hook to `src/useId.ts` and simplify the previous `useId` util? I'm mostly thinking about reusing `generateId` as prefix, which would ensure unique ids in the case where we have multiple React apps using multiple TextField components on one page. I found an article about that here: https://hetdesai03.medium.com/a-complete-guide-to-useid-hook-in-react-18-22119ecfd87f Wdyt?

imprashast: I think that old `useId` solution exists from react 16 days. This hook was added in React 18. 

Notice also our `useId` hook used `useLayoutEffect` which is a performance pitfall, but it was common to use during React 16 days.

I'd say, we should replace instances of our `useId` with React's `useId` but maybe not in this PR.

BalbinaK: I see. But what about `generateId` prefix util that would increase uniqueness of the id? If React's `useId` only generates `r0`, `r1` strings, the chance of multiple TextFields across multiple React apps having the same id is higher. That's why I was thinking that we could maybe prefix those ids with whatever `generateId` returns, the same way it was done in the utils.

imprashast: Hmm, the docs says that react will always generate unique id's for each useId within the vdom. Do you think there might be a situation where this won't hold? 

> multiple TextFields across multiple React apps
Do you mean in case of podlets?

BalbinaK: > Do you mean in case of podlets?
Yes, I was thinking about that :) If there were multiple React apps on a layout/page, served as podlets, then we could theoretically have multiple text fields with the same id, e.g. `r0`.

imprashast: Okay, so I have updated our useId hook and tested it with other components :)

vsandvold:
Hey, thanks for fixing this problem so quickly!

You're right about multiple React apps on the same page getting the same IDs, unless each app is assigned an unique prefix. The React doc mentions a build-in solution for this, which is kind of obscure and indirect:
- https://react.dev/reference/react/useId#specifying-a-shared-prefix-for-all-generated-ids
- https://react.dev/reference/react/useId#using-the-same-id-prefix-on-the-client-and-the-server

Your solution would be much better if it automatically assigns an unique prefix, and just works.

Perhaps I'm misunderstanding something, but it looks like the `generateId()` function includes  a random number in the ID. Wouldn't this number be different from every function call, and therefore also different on the server and client? I would not be surprised if you have already checked that this works for SSR, so I'm just asking to be sure :-)

packages/textfield/src/component.tsx
```diff
@@ -28,9 +28,8 @@ export const TextField = forwardRef<HTMLInputElement, TextFieldProps>(
     } = props;
 
     activateI18n(enMessages, nbMessages, fiMessages);
-
+
```
BalbinaK, line 31: Screenshot 2024-01-24 at 10 53 15

BalbinaK (approved):
💪

PR #202, opened 2024-02-06
fix: bump deps

imprashast:
Just bumping some deps, most importantly `scroll-doctor`

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,14 +107,14 @@
   },
   "dependencies": {
     "@chbphone55/classnames": "^2.0.0",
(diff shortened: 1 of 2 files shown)


BalbinaK (commented):
Just checkin - any changes that are required after @warp-ds/css 1.8.0 update?

imprashast:
I think it's only the addition of alpha channel stuff, but @magnuh knows more

magnuh:
css 1.8.0 includes the rgb token updates and a lot of deprecations... shouldn't need any changes. 🤞 😄

PR #211, opened 2024-02-15
fix: Update warp-ds/css dependency

magnuh:


BalbinaK (approved):
In my opinion that makes sense, but just wanted to showcase that with this update, the disabled active track will be as thick as the enabled one.

Before:
Screenshot 2024-02-15 at 13 29 40

After:
Screenshot 2024-02-15 at 13 29 25

PR #212, opened 2024-02-16
feat: Individual imports

imprashast:
Adding support to be able to import individual components to make the client bundle tree shake properly.


build.mjs
```diff
@@ -0,0 +1,108 @@
+import esbuild from "esbuild";
+import { glob } from "glob";
+import fs from 'fs';
+import ts from 'typescript';
+
+const components = glob.sync("packages/**/src/index.ts");
+const indexPath = "packages/index.ts";
```
BalbinaK, line 7: Should we expose dead-toggle, clickable, affix and so on from _helpers/ folder as separate components, too?

imprashast: Good point, didn't know that 👍

imprashast: These helper components needs to be moved out of _helpers and should follow the same pattern (packages/**/src) for typescript to work properly. I'll create a new task for that and do those in the next PR.

BalbinaK (approved):
LGTM

PR #214, opened 2024-02-22
test(pagination): add tests to get 100% coverage

BalbinaK:
## Description
We are aiming at adding unit tests to all components and having test coverage at 100%. This PR does that for Pagination component.

## What's changed
- added tests erroneous states
- removed redundant safeguards for prop values lower than 1 (`usePagination()` hook always sets page number values to at least 1)
- set a `skipFull` flag to `true` to not show test coverage of components that are already at 100% (reduces noise)

### DIFF
--- a/packages/pagination/src/CurrentPage.tsx
+++ b/packages/pagination/src/CurrentPage.tsx
@@ -15,10 +15,6 @@ const CurrentPage = React.forwardRef<HTMLDivElement, CurrentPageProps>(
   ({ children, className, ...props }, ref) => {
     const { currentPage } = usePagination();
 
-    if (currentPage < 1) {
-      return null;
-    }
-
     const label =
       children ??
       i18n._({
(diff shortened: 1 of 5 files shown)


felicia-haggqvist (approved):
Looks good! :) However, I get these ts warnings for .`toBeInTheDocument()` and `.toHaveTextContent()` in the PaginationTest.tsx file: "Property 'toHaveTextContent' does not exist on type 'Assertion<HTMLElement>'." Not sure if you also get those warnings?

BalbinaK:
> Looks good! :) However, I get these ts warnings for .`toBeInTheDocument()` and `.toHaveTextContent()` in the PaginationTest.tsx file: "Property 'toBeInTheDocument'/'toHaveTextContent' does not exist on type 'Assertion'." Not sure if you also get those warnings?

I did before but now they disappeared, maybe after I restarted TS server at some point?

PR #215, opened 2024-02-23
feat: Add more features on how attention component is positioned

felicia-haggqvist:
**Part of fixes for Jira-tickets:** [WARP-432](https://nmp-jira.atlassian.net/browse/WARP-432), [WARP-341](https://nmp-jira.atlassian.net/browse/WARP-341) & [WARP-373](https://nmp-jira.atlassian.net/browse/WARP-373)

- Add `distance`, `skidding`, `flip`, `fallbackPlacements` props to Attention component
- Update `pointingAtDirection` to also account for the new placements: `top-start`, `top-end`, `right-start`, `right-end`, `left-start`, `left-end`, `bottom-start`, `bottom-end`
- Update `arrowDirectionClassname` to also handle when placement includes either `-start` or `-end` 
- Add a `useEffect` to handle when to call `autoUpdatePosition` that will add event-listeners for on scroll and on resize and calls `recompute`. When `cleanup` is returned, the event-listeners will be removed. 
- Calculation of arrowRotation has been removed from this repo and is now instead calculated in the `warp-ds/core` repo.
- Update Attention examples in `Attention.stories.tsx` to also show how `distance`, `skidding`, `flip`, `fallbackPlacements` and the new placements can be used. 
- Add `AttentionTest.tsx` and add tests for Attention component
- Add `Attention.ts` as a helper file for Attention component
- Bump version of `@warp-ds/core` to use latest changes


packages/attention/stories/Attention.stories.tsx
```diff
@@ -61,14 +62,16 @@ export function Highlight() {
...
           </Box>
         </div>
         <Attention 
           highlight
-          placement='top'
+          placement='right-start'
           isShowing={show}
           targetEl={targetEl}
+          flip={true}
+          fallbackPlacements={['bottom-end']}
```
BalbinaK, line 74: Should bottom-end show the arrow on the right side maybe? This is what I see when the fallback is executed:
Screenshot 2024-03-08 at 13 29 40

packages/attention/src/component.tsx
```diff
@@ -245,33 +317,4 @@ export function Attention(props: AttentionProps) {
...
-  return (
-    <div
-      ref={ref}
-      className={arrowClasses}
-      style={{
-        // TW doesn't let us specify exactly one corner, only whole sides
-        borderTopLeftRadius: '4px',
-        zIndex: 1,
-        // border alignment is off by a fraction of a pixel, this fixes it
-        [`margin${arrowDirectionClassname(arrowDirection)}`]: '-0.5px',
```
BalbinaK, line 272: Removing negative margin may result in borders not aligning:
Screenshot 2024-03-08 at 13 36 20

packages/attention/src/component.tsx
```diff
@@ -176,24 +200,72 @@ export function Attention(props: AttentionProps) {
...
+       return cleanup
+    } 
+  }, [targetEl, isShowing, attentionEl, flip])
+  
+  
+  
+
+  
+
+  const Arrow = forwardRef<HTMLDivElement, ArrowProps>(
```
BalbinaK, line 247: Was it necessary to move `Arrow` component and `arrowDirectionClassname` inside the Attention component scope?

packages/attention/stories/Attention.stories.tsx
```diff
@@ -163,8 +162,8 @@ export function Tooltip() {
 
 export function Popover() {
```
BalbinaK, line 163: Should we add `flip` to this story's Attention, so the component follows the button when screen is resized? Otherwise `autoUpdatePosition` won't trigger because we check if `isShowing && flip && targetEl && attentionEl` in the respective useEffect. Or maybe `flip` missing shouldn't prevent the auto update anyway?

Here's how this story behaves now:
popover

packages/attention/src/props.ts
```diff
@@ -93,4 +96,22 @@ export type AttentionProps = {
    * Handler that is called when the close button is clicked
    */
   onDismiss?: () => void;
+
+  /**
+   * The distance from which to offset the attentionEl from the targetEl vertically
+   */
+distance?: number;
```
BalbinaK, line 103: We really should add @warp-ds/eslint-config and husky integration to this repo soon 😅 There are some changes in this PR that made unnecessary diff noise. Maybe we could prioritize [a task for fixing it](https://nmp-jira.atlassian.net/browse/WARP-446) in the upcoming weeks! 🤞

BalbinaK (commented):
Really impressive work on this one! 🎉 Just few small comments.

packages/attention/src/component.tsx
felicia-haggqvist: hmm true! I'll test to revert the removed code! :)

packages/attention/src/component.tsx
felicia-haggqvist: Yes! If I remember correctly, I noticed that the arrow's ref was not being set when it was outside of the attention element.

felicia-haggqvist: Hmm If I remember correctly, I moved them into the attention component because I had some issues with the arrowRef not updating correctly. But I see now that it seems to work fine also being outside of the attention component, so I can revert this move! :)

packages/attention/src/component.tsx
felicia-haggqvist: I have now reverted this change, but I decided to move `arrowDirectionClassname` into the core-library instead, since that one is exactly the same in react, vue and elements. And I have also moved in this `[`margin${arrowDirectionClassname(arrowDirection)}`]: '-0.5px',` to the core-library, so we just need to add it once! :)

packages/attention/src/component.tsx
felicia-haggqvist: Reverted moving in `Arrow` and have also removed `arrowDirectionClassname` and use instead the `arrowDirectionClassname` that is coming from the core-library. Updated in this commit: [d147451](https://github.com/warp-ds/react/pull/215/commits/d1474513cb2bcb3efc7dc78c9c56d9fea6ede062)

packages/attention/src/props.ts
felicia-haggqvist: Agree! 👍

packages/attention/stories/Attention.stories.tsx
felicia-haggqvist: Updated in this commit: [d22f435](https://github.com/warp-ds/react/pull/215/commits/d22f435b3891882e6751998ae83945e343665946)

packages/attention/stories/Attention.stories.tsx
felicia-haggqvist: Sure! That makes more sense! :)

packages/attention/stories/Attention.stories.tsx
BalbinaK: Just wondering, why do we need to check if `flip` is truthy when running autoUpdate? Would it break something if we run it with `flip` equal `false` or `undefined`?

felicia-haggqvist: I noticed that if you run `autoUpdate` and don't specify how the `attentionEl` should reposition to stay in view as long as possible, then if you, for example set the `placement` to `right` or `left` and then start scrolling down, the `attentionEl` will continue to stay in view, even when the `targetEl` is no longer in view. At first, I thought that was a bug, but I then found out that the `[flip()](https://floating-ui.com/docs/flip)` function in floating-ui gives you the option to specify what should happen if the attentionEl doesn't fit anymore on the screen. So if you keep calling `autoUpdate` it will keep on recalculating the `attentionEl`'s position to keep it in view, but it doesn't know which `placement` we want it to fallback to if the original `placement` doesn't fit anymore, so it stays with the original `placement`. This isn't an issue if the `attentionEl` has `placement` `top` or `bottom`, only on the sides.

felicia-haggqvist: Although, I just found now that you can also use floating-ui's [autoPlacement()](https://floating-ui.com/docs/autoPlacement), which we could use when flip is not enabled. `flip()` gives the user more options to configure how the attentionEl should reposition itself when there is no space, but if we want to call `autoUpdate` even when flip is not enabled, then we could call `autoPlacement`.

BalbinaK: I think that would be nice because otherwise users might not know they need to specify `flip` to get any auto updating. They would need to check the Attention logic to see that `flip` is needed in their setup for that. But I don't know how much work that requires, so maybe setting `flip` to `true` by default IF other conditions are met would be an alternative? It's hard to remember all conditions here so ignore me if my suggestion doesn't make any sense :D

felicia-haggqvist: Yes I agree! I'll add `autoPlacement` and make sure that it will be enabled if the user hasn't enabled `flip`. I tried to set default value to true for `flip` but then I found issues with web components not working if you want to set `flip`to false in the markup, see this [issue](https://stackoverflow.com/questions/60678891/how-can-i-change-a-boolean-property-in-lit-element-to-hide-content-on-my-compone). So that's why I had to set `flip` as default `false`.

BalbinaK: Ah, great!

packages/attention/stories/Attention.stories.tsx
```diff
@@ -213,5 +214,68 @@ export function Popover() {
...
+        fallbackPlacements={['bottom']}
+        distance={2}
+        skidding={12}
+        targetEl={targetEl}
+        isShowing={show}
+      >
+        <ul className='bg-white w-full text-center'>
+          <li
+            tabIndex={0}
+            className='p-2 px-40 hover:bg-blue-200 hover:text-blue-500 cursor-pointer'
```
BalbinaK, line 266: `text-blue-500` and `bg-blue-200` won't work in Warp. It was probably copied from another example here that wasn't fixed after moving over from Fabric. These classes + `cursor-pointer` together with `tabIndex={0}` should probably be removed in order to show plain text, or you could add link elements as children to the `li` elements:
```
  <li className='p-2 px-40' >
    <a href='#/url1'>
      Link 1
    </a> 
  </li>
  <li className='p-2 px-40'>
    <a href='#/url2'>
      Link 2
    </a>
  </li>
```

packages/attention/src/component.tsx
```diff
@@ -12,43 +13,47 @@ import { messages as nbMessages } from './locales/nb/messages.mjs'
...
-    placement,
+    placement = 'bottom',
     targetEl,
     className,
     canClose,
     onDismiss,
+    distance = 8,
+    skidding = 0,
+    flip = false,
+    fallbackPlacements = undefined,
```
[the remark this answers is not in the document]
felicia-haggqvist: Oh you are absolutely right! 😆 Removed default value in this commit: [985956f](https://github.com/warp-ds/react/pull/215/commits/985956f2335bf1c8716a81b36612fb703b0ff1b4)

packages/attention/stories/Attention.stories.tsx
```diff
@@ -213,5 +214,68 @@ export function Popover() {
...
+    }
+  })
+
+  return (
+  <div className='flex items-center justify-center'>
+    <div ref={containerRef}>
+      <Button
+        small
+        aria-expanded={show}
+        aria-controls='popover-icon-target-el-attention-example'
```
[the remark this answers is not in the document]
felicia-haggqvist: Oh good catch! I had only copied the popover-example, and I see now that that one is also missing an id. Added id for both popover and popover icon target el examples in this commit: [336d989](https://github.com/warp-ds/react/pull/215/commits/336d989173c390baf3465d3e20994a08453842ea)

packages/attention/stories/Attention.stories.tsx
felicia-haggqvist: Nice catch! 😄 This has also been copied from the popover example. I have removed these classes + tabIndex from both popover and popover with icon as targetEl examples in this commit: [0985eb4](https://github.com/warp-ds/react/pull/215/commits/0985eb49b76424114e9d76f4cad8d3f695c8079b)

tests/components/AttentionTest.tsx
```diff
@@ -0,0 +1,77 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import { describe, expect, it, } from 'vitest';
+import { Attention } from '../../packages/attention/src/component';
+
+
+describe('Attention component', () => {
```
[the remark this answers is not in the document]
felicia-haggqvist: Added several more tests: [tests/components/AttentionTest.tsx](https://github.com/warp-ds/react/blob/6ca1c32e2e120a7d0daeb13bc747bc87f4a32d8a/tests/components/AttentionTest.tsx)

packages/attention/stories/Attention.stories.tsx
```diff
@@ -196,22 +197,83 @@ export function Popover() {
...
+    return () => {
+      document.removeEventListener('mousedown', onBlurHandler)
+    }
+  })
+
+  return (
+  <div className='flex items-center justify-center'>
+    <div ref={containerRef}>
+      <Button
+        small
```
BalbinaK, line 241: Not sure if it's a bug or not, but the small button's vertical padding cuts the icon 🤔 
Screenshot 2024-04-09 at 10 49 17
Maybe until we figure out what's wrong, you could remove the `small` prop? 😄 It looks good in the regular variant:
Screenshot 2024-04-09 at 10 49 27

felicia-haggqvist: ah nice catch! I can remove the `small` prop!

felicia-haggqvist: Hmm although, I don't get that the small button's vertical padding cuts the icon: 
Skärmavbild 2024-04-09 kl  11 02 18
🤔

felicia-haggqvist: Updated in this commit: [70eb1e0](https://github.com/warp-ds/react/pull/215/commits/70eb1e0fb7344ce19cfc1c4970caa8ab0ff95cfd)

BalbinaK (approved):
LGTM

PR #216, opened 2024-02-26
fix(toggle): only allow label to be 'small' when type is 'radio-button'

BalbinaK:
According to our design, only Toggle of type 'radio-button' should be possible to made "small". There are no small checkbox or radio Toggles in our library.


BalbinaK:
I haven't yet looked into code, just tested locally with changes from css and core, and I noticed that the "edge" variants of arrow position look a bit different than what I found in [Figma](https://www.figma.com/file/oHBCzDdJxHQ6fmFLYWUltf/Warp---Components-2.0?node-id=1308%3A38513&mode=dev).

Figma:
Figma sketch of Attention popover with an arrow in the top-right corner with some right padding

Attention with 'bottom-start' placement:
Attention popover with an arrow in the top-right corner without any right padding

BalbinaK:
Closing as this was fixed in https://github.com/warp-ds/react/pull/222

PR #218, opened 2024-02-29
fix(useId): return fallback ID for React v17

vsandvold:
With this PR, the provided fallback ID is also returned when `React.useId` is not supported. This is the case for React v17.


BalbinaK (approved):
Thanks for the contribution! 
I think the storybook preview job would fail with external contributors, but I've tested that nothing is broken there, so we should be good to go! Just keep in mind that this change will for some time only be available on `next` version of the package.

PR #223, opened 2024-03-08
fix(dead-toggle): add 'name' prop

BalbinaK:
## Description
Fixes [WARP-514](https://nmp-jira.atlassian.net/browse/WARP-514)

Fixes behaviour of a checked `DeadToggle` of type `radio` when multiple groups of `DeadToggles` are rendered on the same page.

| Before | After |
| ----- | ----- |
| Screenshot of 2 groups of cards with radio dead-toggles, both groups have 1 selected card, but in the first group the radio element is not checked | Screenshot of 2 groups of cards with radio dead-toggles, both groups have 1 selected card, and both radio elements are checked

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
     "@chbphone55/classnames": "2.0.0",
     "@lingui/core": "4.7.0",
     "@warp-ds/core": "1.0.2",
-    "@warp-ds/css": "1.8.4",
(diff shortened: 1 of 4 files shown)


felicia-haggqvist (approved):
LGTM! :) Nice job! 🎉

PR #233, opened 2024-04-22
feat: add more functionality to attention component

felicia-haggqvist:
**Part of fixes for Jira-tickets:** [WARP-432](https://nmp-jira.atlassian.net/browse/WARP-432), [WARP-341](https://nmp-jira.atlassian.net/browse/WARP-341) & [WARP-373](https://nmp-jira.atlassian.net/browse/WARP-373)

- Add `crossAxis` prop
- Only render attention component when `targetEl` is not undefined or when it is a `callout`
- Bump to latest version of `@warp-ds/core` dependency
- Add additional tests to `AttentionTest.tsx` to test for when `targetEl`is undefined or when `callout` is either true or false.


packages/attention/src/component.tsx
```diff
@@ -137,54 +141,58 @@ export function Attention(props: AttentionProps) {
...
-            onClick={onDismiss}
-            onKeyDown={(event) => {
-              if (!props.onDismiss) return
-              if (event.key === 'Escape') {
-                props.onDismiss()
-              }
-            }}
-            className={ccAttention.closeBtn}
+    <>
+      {(props.callout || (props.targetEl !== undefined && !props.callout)) && (
```
BalbinaK, line 145: What do you think about doing the opposite here, too, and returning `null` if `!props.callout && props.targetEl === undefined`?

felicia-haggqvist: Yes, I was thinking of doing the same here as well! 🎉 However, not sure if I should make the same changes for the vue attention component: https://github.com/warp-ds/vue/pull/161/files#diff-73fbe20fc6b40d2120b1a5e414f088ed51cea513e1eb9745402ee02a639b22b0R220

BalbinaK: Yeah, sounds like Vue should stay as is with the v-if syntax.

felicia-haggqvist: Updated in this commit: [5c799d2](https://github.com/warp-ds/react/pull/233/commits/5c799d27f0224b55b99cfc94b1f81b924506a6d8)

packages/attention/src/component.tsx
```diff
@@ -137,54 +141,58 @@ export function Attention(props: AttentionProps) {
...
-            )}
-            onClick={onDismiss}
-            onKeyDown={(event) => {
-              if (!props.onDismiss) return
-              if (event.key === 'Escape') {
-                props.onDismiss()
-              }
-            }}
-            className={ccAttention.closeBtn}
+    <>
```
BalbinaK, line 144: I don't think we need this? It would at least make the diff much smaller :D

felicia-haggqvist: I tried to remove `<><>` but I get this error then: `This expression is not callable. Type '{}' has no call signatures.ts. Property assignment expected.` 😕

BalbinaK: The curly braces (`{}`) can be removed from the return statement when you remove `<>`, and that should fix it :)

felicia-haggqvist: Updated in this commit: [7039695](https://github.com/warp-ds/react/pull/233/commits/7039695d139f54d1f290b4157e5ee0842a166128)

BalbinaK (approved):

PR #234, opened 2024-04-24
chore: Add some tests

imprashast:
Adding some tests for some components. This is part 1 of many


tests/components/AlertTest.tsx
magnuh, line 1: Do we need the "Test" suffix on all these test files when they are in the test folder? 😅

imprashast: Hmm, Good question. It just makes it easier to search in IDE

magnuh: Bad IDE then? 😜

tests/components/TextfieldTest.tsx
```diff
@@ -0,0 +1,65 @@
...
+  };
+
+  it('renders correctly', () => {
+    render(<TextField {...defaultProps} />);
+    expect(screen.getByText(defaultProps.label)).toBeInTheDocument();
+  });
+
+  it('renders input field with correct class names', () => {
+    render(<TextField {...defaultProps} />);
+    expect(screen.getByRole('textbox')).toHaveClass('block text-m mb-0 leading-m s-text s-bg s-border hover:s-border-hover active:s-border-selected rounded-4 py-12 px-8 block border-1 w-full focusable focus:[--w-outline-offset:-2px] caret-current');
```
BalbinaK, line 19: In Toggle tests I used imports of component classes from @warp-ds/css instead, and I'm wondering if it was correct, seeing that you used copy-pasted classes here. Wdyt?

imprashast: My reason for this is when someone will do the cleanup, this test will fail which is what the test should do. And then that someone should fix the test.

if it was just importing the css from component classes, the test will never fail and then no reason to even have the test :)

imprashast: I think all tests should follow this because that'll help with the cleanup and also being able to see what classes are actually applying for each variant/prop combination :)

BalbinaK: Makes sense!

magnuh: Using hard coded classes in this way would make us need to update the tests when updating the css dependency, if any classes has changed... Maybe a good way to spot potential problems? 🤷‍♂️ 🙂

PR #237, opened 2024-04-30
chore: add more tests

imprashast:
What's left after this, Tests for:

1. tabs
2. combobox

Coverage at 77.56%


packages/_helpers/expand-transition.tsx
```diff
@@ -46,6 +46,7 @@ export function ExpandTransition({
       className={initialStyle}
       ref={expandableRef}
       aria-hidden={!show ? true : undefined}
+      data-testid="expand-transition"
```
BalbinaK, line 49: Maybe it was already discussed before, but I'm wondering if we should hardcode the data-testid string, or if it makes sense to have a `dataTestId` prop and maybe default it this way (or with a `w-` prefix?). Then both we and whoever uses our components should be able to rely on the `getByTestId()` even when multiple instances of our components are rendered. Here's an article about that: https://medium.com/@kavianrabbani/the-hidden-danger-of-hardcoded-data-testid-in-testing-8ec9e671955e

imprashast: Ideally our consumers shouldn't use our testid's for testing but if they do, we cannot stop them :)

BalbinaK: > Ideally our consumers shouldn't use our testid's for testing but if they do, we cannot stop them :)

Oh, I thought that us having `data-testid` in our components was supposed to be an added value for our users. But I guess we can always add a prop like `dataTestId` later if there's a well-grounded need for that 👍

tests/components/ExpandableTest.tsx
```diff
@@ -0,0 +1,79 @@
...
+    const chevronDown = screen.getByTitle('Downward arrow');
+    expect(chevronDown).toBeInTheDocument();
+    fireEvent.click(screen.getByRole('button'));
+    act(() => {
+        /* fire events that update state */
+        vi.advanceTimersByTime(250);
+        const chevronUp = screen.getByTitle('Upward arrow');
+        expect(chevronUp).toBeInTheDocument();
+    });
+  });
```
BalbinaK, line 70: Maybe I'm misunderstanding something, but if this test doesn't work, what is its purpose here?

tests/components/ExpandableTest.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  it('renders animated expansion correctly', () => {
+    render(<Expandable title="Animated Test" animated>Animated Content</Expandable>);
+    const content = screen.getByTestId('expand-transition');
+    //expect(content).not.toHaveClass('overflow-hidden h-0');
```
BalbinaK, line 75: Should this be commented out?

tests/components/ModalTest.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import { render, fireEvent, waitFor, screen } from '@testing-library/react';
+import { Modal } from '../../packages/modal/src/component';
+
+describe('Modal', () => {
```
BalbinaK, line 5: Should we test component classes strings in all Warp components, too, or does it not make sense?

tests/components/StepsTest.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  it('renders correct aria-current attribute', () => {
+    render(<Step active>Step content</Step>);
+    const stepDot = screen.getByRole('img');
+    expect(stepDot).toHaveAttribute('aria-current', 'step');
+  });
+
+  it('renders IconCheck16 when completed prop is true', () => {
+    render(<Step completed>Step content</Step>);
+    const icon = screen.getByRole('img');
```
BalbinaK, line 62: Isn't the div with role="img" always present with a `Step` is rendered? How else could we check if an icon is rendered?

tests/components/SwitchTest.tsx
```diff
@@ -0,0 +1,63 @@
...
+  it('updates the handle element class name based on the value prop', () => {
+    const handleSelected = 'absolute transform-gpu h-16 w-16 top-4 left-4 rounded-full transition-transform s-bg shadow-s translate-x-20';
+    const { rerender } = render(<Switch value={false} />);
+    expect(screen.getByTestId('handle')).not.toHaveClass(handleSelected);
+    rerender(<Switch value />);
+    expect(screen.getByTestId('handle')).toHaveClass(handleSelected);
+  });
+
+  it('passes through additional props to the button element', () => {
+    render(<Switch data-test="my-test" />);
```
BalbinaK, line 60: Will it also override our data-testid value if you pass `data-testid="my-test"` here? :D

BalbinaK (commented):
Supercool stuff! I left few minor comments below but overall looks great!

tests/components/ExpandableTest.tsx
imprashast: I'll remove

tests/components/SwitchTest.tsx
imprashast: Hmm, since we set rest, it just might. Should we filter stuff in rest, that's probably a different task :)

tests/components/StepsTest.tsx
imprashast: Good point. I didn't realize that. Maybe this should be improved in the next PR

tests/components/ModalTest.tsx
BalbinaK: I thought handling tests before the cleanup tasks was supposed to help the reviewers see which classes were changed as a result of the cleanup 😊 I also saw you added similar tests in other components, so I was wondering if that should be applied to all components or only to some. But sure, this can be added as part of the classes cleanup task 👍

tests/components/StepsTest.tsx
BalbinaK: Not sure which next PR you have in mind, but would it solve it if we added a `data-testid="completed-icon"` to the `<IconCheck16 />` component, and then changed the test to something like this?
```
it('renders IconCheck16 when completed prop is true', () => {
    render(<Step completed>Step content</Step>);
    const icon = screen.getByTestId('completed-icon');
    expect(icon).toBeInTheDocument();
  });
  ```

tests/components/ModalTest.tsx
imprashast: Modal discussion with designers is ongoing, so if the paddings are changed, we should revisit this before doing a component cleanup :)

PR #238, opened 2024-05-02
feat: add warp ds eslint config

felicia-haggqvist:
Fixes Jira issue: [WARP-520](https://nmp-jira.atlassian.net/browse/WARP-520)

- Add eslint configuration and workflow
- lint all code using `pnpm lint`


vite.config.js
```diff
@@ -22,23 +22,7 @@ export default function config(env) {
...
-          "lingui.config.ts",
-          "@types",
-          "tests/",
-          "packages/**/src/props.*",
-          "packages/**/src/index.*",
-          "packages/index.ts",
-          "build.js",
-          "esbuild.js",
-        ],
+        exclude: ['**.json', 'dev/**', 'storybook-static/**', '.storybook/**', 'packages/**/stories', 'packages/**/src/locales', 'esbuild.mjs', 'lingui.config.ts', '@types', 'tests/', 'packages/**/src/props.*', 'packages/**/src/index.*', 'packages/index.ts', 'build.js', 'esbuild.js'],
```
BalbinaK, line 25: I can't wrap my head around why this wrapped content was formated to be one line, while in other repos that didn't happen. We had the same setup e.g. in @warp-ds/vue, which extends @warp-ds/eslint-config. I see that in our Warp config we specified `printWidth` for prettier to 999 ([code](https://github.com/warp-ds/eslint-config/blob/master/index.js#L52C4-L58C7)), which sounds like the setting responsible for this, but it only seems to apply in this repo. I'm also wondering, is `printWidth: 999` rule for prettier something we would like to have? @warp-ds/warp-core-team

felicia-haggqvist: Yeah I don't know why it didn't format the same way as for vue 🙈 😕

imprashast: It shouldn't exceed 140 `"printWidth": 140,`
This would work on most retina display macbooks without issues

imprashast: It also might be because of `"plugin:prettier/recommended"` in the eslintrc 🤔

felicia-haggqvist: No it did that formatting before I added `"plugin:prettier/recommended"` in the eslintrc

felicia-haggqvist: Although it looks like it had atleast formatted the `exclude` the same way in the `vite.config.js`file for vue. 🤔 See [here](https://github.com/warp-ds/vue/pull/152/files#diff-58e6f63d87181b1c6a8cb6e5f1691df04aa32854456efcd52ca71c8541375d26R27)

BalbinaK: Isn't it the `"printWidth": 999` from [@warp-ds/eslint-config](https://github.com/warp-ds/eslint-config/blob/master/index.js#L52C4-L58C7) that is kicking in here (and then it didn't in the other repos?)? When I added `printWidth: 150` as a separate rule in `.eslintrc` here, it worked fine. 999 characters is an eternity in the world of an IDE, so I'm not sure why it was set to that in the first place 😅

felicia-haggqvist: But it reacted the same way in vue as in react (see my previous comment). So it did kick in. Talked to @imprashast and we talked about that I should update the @warp-ds/eslint-config to have `printWidth: 140` :)

.github/workflows/lint.pr.yaml
```diff
@@ -0,0 +1,21 @@
...
+    - name: Checkout repository
+      uses: actions/checkout@v4
+    - name: Setup Node.js
+      uses: actions/setup-node@v4
+      with:
+        node-version: lts/*
+    - name: Install pnpm and dependencies
+      uses: pnpm/action-setup@v2
+      with:
+        version: 8
```
BalbinaK, line 18:
```suggestion
        version: 9
```

BalbinaK (commented):
Just one comment and we should be good to go!

BalbinaK (approved):

PR #239, opened 2024-05-03
fix(textfield/textarea): clean up classes

BalbinaK:
## Description
Use cleaned up input component classes to avoid buggy style overriding when utility classes styling the same CSS properties were applied to the same element

## Changes included
- Update @warp-ds/uno to 1.12.0 in order to handle padding classes with arbitrary values that contain CSS variables (e.g. `pl-[var(--w-prefix-width,_40px)]`
- Use the cleaned up input component classes to avoid style overriding
- Set the pnpm version to v9 in github workflows to avoid lockfile issues after updating pnpm locally

## Testing
Tested with TextField and TextArea.
Gif representing TextField in different states behaving and looking the same as before

### DIFF
--- a/.github/workflows/i18n.yml
+++ b/.github/workflows/i18n.yml
@@ -15,6 +15,7 @@ jobs:
       - name: Install pnpm and dependencies
         uses: pnpm/action-setup@v2
         with:
+          version: 9
           run_install: true
       - name: Validate extracted messages
         run: pnpm messages:compile
--- a/.github/workflows/i18n.yml
+++ b/.github/workflows/i18n.yml
@@ -15,7 +15,6 @@ jobs:
       - name: Install pnpm and dependencies
         uses: pnpm/action-setup@v2
         with:
-          version: 8
           run_install: true
       - name: Validate extracted messages
         run: pnpm messages:compile
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -23,7 +23,6 @@ jobs:
       - name: Install pnpm and dependencies
         uses: pnpm/action-setup@v2
         with:
-          version: 8
           run_install: true
       - name: Build storybook
         run: pnpm run build-storybook
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -23,6 +23,7 @@ jobs:
       - name: Install pnpm and dependencies
         uses: pnpm/action-setup@v2
         with:
+          version: 9
           run_install: true
       - name: Build storybook
         run: pnpm run build-storybook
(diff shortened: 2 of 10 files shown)


felicia-haggqvist (approved):
LGTM! :)

Skadefryd (approved):
Nice :)  Looks good to me!

BalbinaK:
@warp-ds/warp-core-team should I maybe set the version of `pnpm` in the Github Action jobs to `v9` instead of removing that setting and adding `"packageManager": "9.1.1"` in `package.json`? The problem is that by default the version will be strictly required when running `pnpm install`, which means we'd need to update that version in `package.json` each time someone updates pnpm locally. There was a suggestion to install [asdf](https://asdf-vm.com/) and align the versions by always running `asdf install` for each repository locally, but I'm thinking that this might be too much hassle to fix an issue that has only happened once a year so far (like any other maintenance task). 

If you agree with updating pnpm version to `v9` in `.github/workflows/`, give a 👍 Condition here is that we all would then need to update `pnpm` locally to v9 with `pnpm add -g pnpm` to avoid huge diffs in `pnpm-lock.yaml` and possible failures in CI jobs.

BalbinaK:
> @warp-ds/warp-core-team should I maybe set the version of pnpm in the Github Action jobs to v9 instead of removing that setting and adding "packageManager": "9.1.1" in package.json? The problem is that by default the version will be strictly required when running pnpm install, which means we'd need to update that version in package.json each time someone updates pnpm locally. There was a suggestion to install [asdf](https://asdf-vm.com/) and align the versions by always running asdf install for each repository locally, but I'm thinking that this might be too much hassle to fix an issue that has only happened once a year so far (like any other maintenance task).
>
> If you agree with updating pnpm version to v9 in .github/workflows/, give a 👍 Condition here is that we all would then need to update pnpm locally to v9 with pnpm add -g pnpm to avoid huge diffs in pnpm-lock.yaml and possible failures in CI jobs.

After having an internal discussion we agreed that I should update the `pnpm` version the github action runner will install to v9 and remove the `packageManager` key from `package.json`. That was done in [ce48f6b](https://github.com/warp-ds/react/pull/239/commits/ce48f6badb580f9aadcdfe2e5d4995bcffd5aa3e) and [ff93e5a](https://github.com/warp-ds/react/pull/239/commits/ff93e5a599bd53484e9755056e784b81560bee72). 
All contributors will need to update pnpm to v9 with `pnpm add -g pnpm` in order to align with the lockfile versions. We've chosen this path with a hope that there will be no more actions required for a long period of time, i.e. until a new major version of pnpm creates a breaking change in the lockfiles, which could cause the CI jobs to fail. If we experience that this sort of maintenance is not working too well over time and we see no better way to solve it, we agreed to reconsider the use of [asdf](https://asdf-vm.com/) or [direnv](https://direnv.net/) in order to align the version of `pnpm` and `node` among all contributors.

felicia-haggqvist (approved):

PR #248, opened 2024-06-11
fix: add default targetEl for callout

felicia-haggqvist:
Fixes Jira [WARP-527](https://nmp-jira.atlassian.net/browse/WARP-527)

- Update to latest version of @warp-ds/core dependency
- Create a default `targetEl` when `callout` is `true` and `targetEl` is `undefined`.
- Add a comment for the reason of creating a default `targetEl`: 
`"useAutoUpdatePosition hook is using @warp-ds/core, which uses Floating-ui's computePosition(). Floating-ui's computePosition() requires a defined targetEl to be able to compute the attentionEl's position and the attentionEl's arrow position. When props.callout is true, we only need computePosition() to calculate the callout's arrow position. So, we create a default targetEl for callout that we can pass to the useAutoUpdatePosition hook, in order to avoid Floating-ui from throwing an error."`
- Add class `ml-8` to the callout stories.
- Fix eslint warnings from test files.

**Tested:**
- Tested in Chrome, Firefox and Safari
- Also tested in Server-side rendered app


packages/attention/src/component.tsx
```diff
@@ -92,7 +93,7 @@ export function Attention(props: AttentionProps) {
         attentionEl.current = v;
       },
       get targetEl() {
-        return targetEl?.current;
+        return targetElRef.current;
```
imprashast, line 96: targetElRef can be null so maybe do `targetElRef?.current` just so the null is handled

tests/components/TextAreaTest.tsx
```diff
@@ -126,6 +126,7 @@ describe('TextArea component', () => {
     const textAreaRef = createRef<HTMLTextAreaElement>();
 
     // Render the TextArea component and pass the ref as forwardRef
+    // eslint-disable-next-line no-unused-vars
     const { container } = render(<TextArea ref={textAreaRef} />);
```
imprashast, line 130: This could just be (without the eslint-disable):

// Render the TextArea component and pass the ref as forwardRef
render(<TextArea ref={textAreaRef} />);

packages/attention/src/component.tsx
felicia-haggqvist: Ah good catch! 👍

imprashast (approved):

PR #249, opened 2024-06-14
fix(select): clean up classes

felicia-haggqvist:
## Description
This PR ensures no classes styling the same CSS properties are being set on the same HTML element in the Select component.

Additionally this PR has added missing props `disabled` and `readOnly` to `props.tsx` file for Select and added a missing Select story for `readOnly`. This PR also adds a `handleKeyDown()` function to disable `space`, `arrowUp` and `arrowDown` for when `readOnly` prop is set to `true`

## How to test
Either link this branch with @warp-ds/css (`fix/clean-up-select-classes` branch) or replace this for @warp-ds/css dependency in package.json: `github:warp-ds/css#component-classes-cleanup` with this  `github:warp-ds/css#fix/clean-up-select-classes`

### DIFF
--- a/packages/select/src/props.tsx
+++ b/packages/select/src/props.tsx
@@ -16,6 +16,12 @@ export type SelectProps = {
   /**  Renders the field in an invalid state. Often paired together with `hint` to provide feedback about the error. */
   invalid?: boolean;
 
+  /**  Renders the field in a disabled state.*/
+  disabled?: boolean;
+
+  /**  Renders the field in a read-only state.*/
+  readonly?: boolean;
+
   /** The content to display as the help text. */
   hint?: React.ReactNode;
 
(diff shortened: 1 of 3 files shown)


imprashast:
Unrelated to the PR, but:

- disabled & readOnly are missing in props.tsx for select. Would be nice to have them added in this PR.
- There is no story for readOnly mode of select. Would be nice to have them added in this PR.

packages/select/src/props.tsx
```diff
@@ -16,6 +16,12 @@ export type SelectProps = {
   /**  Renders the field in an invalid state. Often paired together with `hint` to provide feedback about the error. */
   invalid?: boolean;
 
+  /**  Renders the field in a disabled state.*/
+  disabled?: boolean;
+
+  /**  Renders the field in a read-only state.*/
+  readonly?: boolean;
```
imprashast, line 23: This should be `readOnly`

felicia-haggqvist: Whoopsie! 🙈

imprashast (approved):

BalbinaK (approved):

PR #251, opened 2024-06-19
fix(slider): prevent onChange/onChangeAfter from being called on initial mount

yakupyazicioglu:
This PR prevents `onChange` and `onChangeAfter` from being called immediately after the component mounts.

- Added a ref `hasMounted` to track whether the component has mounted.
- Updated the first useEffect to call `onChange(value)` only if the component has already mounted.
- Updated the second useEffect to call `onChangeAfter(value)` only if the component has already mounted and `sliderPressed` is false.
- Introduced a third useEffect to set the `hasMounted` flag to true after the initial render.
- Ensured that both useEffect hooks respect the `hasMounted` flag, preventing them from running their respective logic on the initial mount.

### DIFF
--- a/packages/slider/src/component.tsx
+++ b/packages/slider/src/component.tsx
@@ -33,15 +33,12 @@ export function Slider({ min = 0, max = 100, ...rest }: SliderProps) {
 
   const updateValue = useCallback(
     (value: number) => {
-      console.log("updateValue", value);
-
       setValue(value);
       onChange && onChange(value);
 
       if (!sliderPressed) {
         onChangeAfter && onChangeAfter(value);
       }
-  useEffect(() => {
     setHasMounted(true);
     },
     [onChange, onChangeAfter, sliderPressed]
@@ -81,22 +78,28 @@ export function Slider({ min = 0, max = 100, ...rest }: SliderProps) {
 
   const { handleKeyDown, handleFocus, handleBlur, handleMouseDown, handleClick, getThumbPosition, getThumbTransform, getShiftedChange } =
     createHandlers({ props: { min, max, ...rest }, sliderState });
+    handleFocus,
+    handleBlur,
+    handleMouseDown,
+    handleClick,
+    getThumbPosition,
+  } = createHandlers({ props: { min, max, ...rest }, sliderState });
 
   const thumbPosition = useMemo(getThumbPosition, [getThumbPosition]);
   const sliderActiveStyle = useMemo(
     () => ({
       left: 0,
-      right: 100 - thumbPosition + '%',
+      right: 100 - thumbPosition + "%",
     }),
     [thumbPosition],
+    [thumbPosition]
   );
 
   const transformValue = useMemo(getThumbTransform, [getThumbTransform]);
   const thumbStyles = useMemo(
     () => ({
-      transform: 'translateX(' + transformValue + 'px)',
+      transform: "translateX(" + transformValue + "px)",
     }),
-    [transformValue],
   );
 
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


vsandvold:
As we talked about offline, there's a slightly different solution that reduces the overall complexity a bit:
```ts
  const [value, setValue] = useState(rest.value);
  // other useState hooks

  // replaces the previous useEffect hooks and hasMounted
  const updateValue = useCallback(
    (v: unknown) => {
      setValue(v);
      onChange?.(v);
      if (!sliderPressed) {
        onChangeAfter?.(v);
      }
    },
    [onChange, onChangeAfter, sliderPressed]
  );

// Replace useValue with updateValue elsewhere
```

packages/slider/src/component.tsx
```diff
@@ -68,6 +77,12 @@ export function Slider({ min = 0, max = 100, ...rest }: SliderProps) {
 
   const { handleKeyDown, handleFocus, handleBlur, handleMouseDown, handleClick, getThumbPosition, getThumbTransform, getShiftedChange } =
     createHandlers({ props: { min, max, ...rest }, sliderState });
+    handleFocus,
+    handleBlur,
+    handleMouseDown,
+    handleClick,
+    getThumbPosition,
+  } = createHandlers({ props: { min, max, ...rest }, sliderState });
```
felicia-haggqvist, line 85: I think the code on lines 80-85 is a duplication of lines 78-79 and should be removed (remove line 80-85). :)

packages/slider/src/component.tsx
```diff
@@ -1,4 +1,10 @@
-import React, { useEffect, useMemo, useRef, useState } from 'react';
```
felicia-haggqvist, line 1: Make sure to run `pnpm lint:check` and `pnpm lint` so that it formats the file correctly. Otherwise you will get a lint-build error :)

packages/slider/src/component.tsx
yakupyazicioglu: I'm having problems with my linter, sorry about that:)

felicia-haggqvist:
I ran `pnpm test` and see that you have two failing tests in `SliderTest.tsx`: 

```
 FAIL  tests/components/SliderTest.tsx > Slider > calls onChange when value changes
AssertionError: expected "spy" to be called 1 times, but got 0 times
 ❯ tests/components/SliderTest.tsx:22:22
     20|     const thumb = screen.getByRole('slider');
     21|     await fireEvent.mouseDown(thumb, { clientX: 50 });
     22|     expect(onChange).toHaveBeenCalledTimes(1);
       |                      ^
     23|   });
     24| 
```

```
 FAIL  tests/components/SliderTest.tsx > Slider > calls onKeyDown when thumb is pressed
AssertionError: expected "spy" to be called 2 times, but got 1 times
 ❯ tests/components/SliderTest.tsx:31:22
     29|     await fireEvent.keyDown(thumb, { key: 'ArrowRight' });
     30|     // not sure why it gets called twice here
     31|     expect(onChange).toHaveBeenCalledTimes(2);
       |                      ^
     32|   });
     33| });
```

felicia-haggqvist:
I noticed when clicking on the slider it will move the slider to 0. Then when I click reset and try to move the slider again it will quickly go to 0 and then follow where my mouse is dragging it. See this recording: 

https://github.com/warp-ds/react/assets/144792260/02ebe75b-b29d-4c22-ba78-4ce3064bfc18

vsandvold:
@felicia-haggqvist Yakup is on holiday now, but I'll take a look at the issues you have found.

vsandvold:
@felicia-haggqvist Please review this PR instead:
https://github.com/warp-ds/react/pull/253

PR #270, opened 2024-08-05
fix(switch): clean up classes

felicia-haggqvist:
## Description
This PR ensures no classes styling the same CSS properties are being set on the same HTML element in the `Switch` component.

**Changes:**
- Refactor classes in `Switch` component
- Remove `switchFocus` since those classes have been moved to `ccSwitchToggle.base` in component-classes
- Add missing `disabled` prop and disabled classes to align with Figma and Vue
- Add additional tests to test for disabled state

## How to test
Either link this branch with @warp-ds/css (`fix/cleanup-switch-toggle-classes` branch) or replace this for @warp-ds/css dependency in package.json: `github:warp-ds/css#component-classes-cleanup` with this  `github:warp-ds/css#fix/cleanup-switch-toggle-classes`

### DIFF
--- a/packages/switch/src/component.tsx
+++ b/packages/switch/src/component.tsx
@@ -5,8 +5,34 @@ import { switchToggle as ccSwitch } from '@warp-ds/css/component-classes';
 
 import { SwitchProps } from './props.js';
 
-export function Switch({ id, value, onClick, 'aria-label': ariaLabel, 'aria-labelledby': ariaLabelledBy, ...attrs }: SwitchProps) {
-  const switchFocus = 'focusable rounded-full';
+export function Switch({
+  id,
+  value,
+  disabled,
+  onClick,
+  'aria-label': ariaLabel,
+  'aria-labelledby': ariaLabelledBy,
+  ...attrs
+}: SwitchProps) {
+  const baseClasses = classNames([ccSwitch.base, disabled && ccSwitch.disabled]);
+
+  const trackClasses = classNames([
+    ccSwitch.track,
+    disabled && ccSwitch.trackDisabled,
+    value && !disabled ? ccSwitch.trackActive : ccSwitch.trackInactive,
+  ]);
+
+  const handleClasses = classNames([
+    ccSwitch.handle,
+    value && ccSwitch.handleSelected,
+    disabled ? ccSwitch.handleDisabled : ccSwitch.handleNotDisabled,
+  ]);
+
+  const handleClick = (e) => {
+    if (!disabled && onClick) {
+      onClick(e);
+    }
+  };
 
   return (
     <div>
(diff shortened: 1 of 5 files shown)


packages/switch/src/component.tsx
```diff
@@ -5,8 +5,34 @@ import { switchToggle as ccSwitch } from '@warp-ds/css/component-classes';
...
+  'aria-label': ariaLabel,
+  'aria-labelledby': ariaLabelledBy,
+  ...attrs
+}: SwitchProps) {
+  const baseClasses = classNames([ccSwitch.base, disabled && ccSwitch.disabled]);
+
+  const trackClasses = classNames([
+    ccSwitch.track,
+    disabled && ccSwitch.trackDisabled,
+    value && !disabled ? ccSwitch.trackActive : ccSwitch.trackInactive,
```
BalbinaK, line 22: Should this condition look similar as you did in a respective [Vue PR](https://github.com/warp-ds/vue/pull/187#discussion_r1706656863)?
```suggestion
    disabled && ccSwitch.trackDisabled,
    !disabled && (value ? ccSwitch.trackActive : ccSwitch.trackInactive)
```

packages/switch/src/component.tsx
```diff
@@ -17,16 +43,14 @@ export function Switch({ id, value, onClick, 'aria-label': ariaLabel, 'aria-labe
         aria-label={ariaLabel}
         aria-labelledby={ariaLabelledBy}
         aria-checked={value}
-        onClick={onClick}
-        className={classNames([ccSwitch.label, switchFocus])}
+        onClick={handleClick}
+        className={baseClasses}
+        tabIndex={disabled ? -1 : 0}
```
BalbinaK, line 48: I don't think this is needed if you have `disabled=true` on a button.

packages/switch/src/component.tsx
```diff
@@ -5,8 +5,34 @@ import { switchToggle as ccSwitch } from '@warp-ds/css/component-classes';
...
+    ccSwitch.handle,
+    value && ccSwitch.handleSelected,
+    disabled ? ccSwitch.handleDisabled : ccSwitch.handleNotDisabled,
+  ]);
+
+  const handleClick = (e) => {
+    if (!disabled && onClick) {
+      onClick(e);
+    }
+  };
```
BalbinaK, line 35: We don't need this check because we have a `pointer-events-none` class set if the Switch is disabled, which makes the button not a target to any pointer events.

BalbinaK (commented):
Nice work! Just a few comments below. We will need to update tech docs regarding the disabled switch: https://warp-ds.github.io/tech-docs/components/switch/#disabled-property

packages/switch/src/component.tsx
felicia-haggqvist: Yes you're right! 😄  I'll change it! 🎉

packages/switch/src/component.tsx
felicia-haggqvist: the `pointer-events-none` seem to only work for the mouse, but you could still navigate with the keyboard and toggle the switch, so that's why I added this handleClick. :)

felicia-haggqvist: Ah right! I'll remove this :)

packages/switch/src/component.tsx
felicia-haggqvist: Updated in this commit: [fb41ad8](https://github.com/warp-ds/react/pull/270/commits/fb41ad85ab2fbdb3561e8692df4e89c07f63fcb0)

packages/switch/src/component.tsx
felicia-haggqvist: Updated in this commit: [fb41ad8](https://github.com/warp-ds/react/pull/270/commits/fb41ad85ab2fbdb3561e8692df4e89c07f63fcb0)

packages/switch/src/component.tsx
felicia-haggqvist: Updated in this commit: [fb41ad8](https://github.com/warp-ds/react/pull/270/commits/fb41ad85ab2fbdb3561e8692df4e89c07f63fcb0)

PR #294, opened 2024-11-19
fix(attention): move role and aria-label to attention-arrow wrapper

felicia-haggqvist:
Fixes Jira issue: [WARP-651](https://nmp-jira.atlassian.net/browse/WARP-651)

**Changes include:**
- Move `role` and `aria-label` from the `div` wrapping the speech bubble to a `div` that is wrapping the attention `Arrow` component


imprashast (approved):

BalbinaK (approved):
Tested with `Control+Option+ArrowLeft/Right` to navigate through non-interactive contents of the Storybook playground and now I hear both aria-label "Speech bubble ..." and then the content of the Attention component being announced 🎉

PR #309, opened 2025-02-25
fix(lingui): Move activateI18n outside of components

oleob:
This will ensure that translations are only loaded once, and will not create an infinite render-loop


imprashast (approved):
LGTM 👍

PR #315, opened 2025-04-04
fix: nullable children inside <Box/>

jakubboesche:
Error scenario:
`<Box>{aBooleanVariable && <div>abc</div>}</Box>`
Error message:
```
error TS2322: Type 'false | Element' is not assignable to type 'Element'.
  Type 'boolean' is not assignable to type 'ReactElement<any, any>'.

175     {!isLightLaunch && <div>abc</div>}

```

### DIFF
--- a/packages/box/src/props.tsx
+++ b/packages/box/src/props.tsx
@@ -4,7 +4,7 @@ export interface BoxProps {
   /**
    * Expand element children
    */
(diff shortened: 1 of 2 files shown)


imprashast:
This can be mitigated by wrapping the children in a React Fragment. Hence closing :)

PR #316, opened 2025-05-09
fix: update label prop types to accept React nodes in multiple compon…

imprashast:
- Build helper function and export them individually
- Change label props to ReactNode


zaramelyan (approved):
lgtm 🎉

PR #318, opened 2025-05-12
fix: update FocusLock import path to use the default export

zaramelyan:
Second attempt to fix FocusLock error.
Generates typescript error `"FocusLock" cannot be used as a JSX component.`,
but according to [docs](https://www.npmjs.com/package/react-focus-lock) it should be valid –
runs in Storybook.


zaramelyan:
Note to self:
check out the initialFocusRef mentioned in [docs](https://warp-ds.github.io/docs/components/modal/) (not currently passed to FocusLock?)

PR #331, opened 2025-10-02
fix: use label for input field preview if available

joms:
Add a utility function to be able to use more rules to calculate the title: only get the label if it's convertible to text

Giving another stab at https://github.com/warp-ds/react/pull/327

### DIFF
--- a/packages/combobox/src/component.tsx
+++ b/packages/combobox/src/component.tsx
@@ -24,7 +24,7 @@ import { messages as fiMessages } from './locales/fi/messages.mjs';
 import { messages as nbMessages } from './locales/nb/messages.mjs';
 import { messages as svMessages } from './locales/sv/messages.mjs';
 import type { ComboboxProps, OptionWithIdAndMatch } from './props.js';
-import { createOptionsWithIdAndMatch, getAriaText } from './utils.js';
+import { createOptionsWithIdAndMatch, getAriaText, getSelectedOptionText } from './utils.js';
 
 const OPTION_CLASS_NAME = 'w-react-combobox-option';
 const MATCH_SEGMENTS_CLASS_NAME = 'w-react-combobox-match';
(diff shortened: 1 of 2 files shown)


zaramelyan (commented):
When I test this on an option that has both `label` and a `value`, the text displayed comes from `label` until selected, and once selected, it changes to the text in `value`. Is this the intended functionality? Do we know if this will have any side effects or change in behavior for current implementations of combobox in other teams?

joms:
> When I test this on an option that has both `label` and a `value`, the text displayed comes from `label` until selected, and once selected, it changes to the text in `value`. Is this the intended functionality?

To me that makes sense: It should show the "temporary value" as a preview, and when the permanent value is set, that is what should be shown until it has a new temporary value.

> Do we know if this will have any side effects or change in behavior for current implementations of combobox in other teams?

Hard to say, but I would say that this shouldn't cause any big surprises, as it just adheres to using the `label` in stead of the `value` in a preview mode. And that does only apply if `label` is actually set.

Skadefryd:
Sorry its very easy to misread the intent here :D    The combobox should never show the value to the end user, only the more humanized Label, (unless there only is value and no label...... If that is even a usecase). When i read this fix, it seams like the data will be shown to the user on keyb nav, but when you hit enter its swapped for the LAbel , If that is so, please explain the usecase for this :)      Above it seams like you are talking about 2 different things. Zara is talking about showing the label OR value to the user, while you seam to answer with the pattern for when to visually update the input field, both of you seam correct to me , just not talking about the same thing ?  :D

joms:
> When i read this fix, it seams like the data will be shown to the user on keyb nav, but when you hit enter its swapped for the LAbel , If that is so, please explain the usecase for this :) 

In the current implementation, the label is ignored on keyboard navigation. This fix actually shows the label while navigating with keyboard.

However, the current implementation has a mode while navigating where it shows the "selected" option before selecting it. That's just existing functionality, the only thing I have done is ensuring that it shows the label. I don't care whether that functionality should stay. But if it should, it should show the label 😅
It is fine if you guys want to discuss whether that should still be a feature of the Combobox, but imo that should be discussed independently of this PR as this solves what I see as a bug in the current release.

On the mobility search page you can see the current problem: https://www.finn.no/mobility/search/car?registration_class=1

image

imprashast (approved):

